PR #5, opened 2015-09-24
[WIP/EXPERIMENTAL] Change implementation to use container

AMHOL:


solnic:
`coerces_from` isn't right, sum-type/composite used to check the input type, it has nothing to do with coercion, it's a type-check, so right now `valid?` logic is different from what it used to be. we don't care about checking input type wrt coercion, the constructor should encapsulate this kind of stuff.

Would it be possible to focus on container first? This PR is too big :/

AMHOL:
:+1:

PR #16, opened 2016-01-12
Passthrough value if type is correct

kwando:
Passthough Date, DateTime and Time in form coercions if the input value already has the correct type.

### DIFF
--- a/lib/dry/data/coercions/form.rb
+++ b/lib/dry/data/coercions/form.rb
@@ -19,18 +19,21 @@ module Dry
         end
 
         def self.to_date(input)
+          return input if input.kind_of?(Date)
           Date.parse(input)
         rescue ArgumentError
           input
         end
 
         def self.to_date_time(input)
+          return input if input.kind_of?(DateTime)
           DateTime.parse(input)
         rescue ArgumentError
           input
         end
 
         def self.to_time(input)
+          return input if input.kind_of?(Time)
           Time.parse(input)
         rescue ArgumentError
           input


solnic:
Why does it already have correct type? The intention of `form.*` types is to use it against input from a web form, which means values are *always* strings.

solnic:
I'm gonna close this as `form` coercions are meant to be used with input where values are always strings. Please tell me about the use case you had and we can think how to solve this for you in a different way.

kwando:
I'm using dry-validation to validate params to a web app and I had a need to add default values for some params.

The simplest way to achieve this is to merge the params hash with a hash of default values.

```ruby
default_values = {'name' => 'kwando'}
input = default_values.merge(params)
form_schema.call(input)
```

If you need to have a date as a default value you need to specify it as a string and then the dry-data parses the date again (wasting cpu cycles).

```ruby
default_values = {'start_at' => Date.today.to_s}
input = default_values.merge(params)
form_schema.call(input).params[:start_at]  # <Date: 2016-02-08 ((2457427j,0s,0n),+0s,2299161j)>
```

1.2 mins later...
Hmm, now that I thinking about it another solution came to mind.
One could specify the attributes that has default values as "optional" in the schema and then merge defaults value hash with ```result.params``` instead... though you would miss validation of your default values that way.

solnic:
@kwando let's wait and see how it's gonna work once I add support for default values to dry-types hashes + use it in dry-validation

kwando:
@solnic no, hurry =) I'm working around it anyway, just wanted to show you my use case.

tiagoefmoraes:
@solnic I have another use case where this would be useful.

I create my Struct from another Struct so the date is already a Date, then I save it to JSON and read it back later, this time the date is a String, as JSON doesn't support the Date type.

The passthrough seems like the easy solution but maybe I should do this in another way. What do you say?

PR #17, opened 2016-01-12
Adding alias method verification for call

jeremyf:
## Updating spec desc to reflect tested method

@326bb2d649d94bb104ed99e96b8283cb48b25e62

The description of the spec stated it was testing #call yet the specs
were instead testing #[]. This change brings the spec description more
inline with other specs for #[].

## Adding alias_method tests for #call

@48707ab41b5a00ae3f49730c6169f33023b20d06

While the method is defined in #call the specs are testing the alias
of #[].

For the projects I'm considering using dry-data, I find the #call idiom to be
more inline with the projects existing idioms. So I want to see specs that
assert #call is part of the public interface and expectation.


solnic:
Makes sense. Thank you.

PR #26, opened 2016-02-06
Dereference optional values.

### LINKED ISSUE #22: Array of structs doesn't work
I thought this should work, but it doesn't...

```ruby
require 'dry-data'

module Types
end

Dry::Data.configure do |config|
  config.namespace = Types
end

Dry::Data.finalize

class Book < Dry::Data::Value
  attribute :name,      Types::Maybe::Strict::String
end

class Shelf < Dry::Data::Value
  attribute :books,     Types::Strict::Array.member(Book)
end

book1 = Book.new(:name => 'A book name')
book2 = Book.new(:name => 'Another book')

puts Shelf.new(:books => [book1, book2]).inspect
```

```
Dry::Data::StructError: [Shelf.new] [#<Book name=Some("A book name")>, #<Book name=Some("Another book")>] (Array) has invalid type for :books
        from /usr/lib/ruby/gems/2.3.0/gems/dry-data-0.5.1/lib/dry/data/struct.rb:36:in `rescue in new'
        from /usr/lib/ruby/gems/2.3.0/gems/dry-data-0.5.1/lib/dry/data/struct.rb:34:in `new'
```

marshall-lee:
Fixes #22.

### DIFF
--- a/lib/dry/types/optional.rb
+++ b/lib/dry/types/optional.rb
@@ -8,7 +8,11 @@ module Dry
       include Builder
 
       def call(input)
-        Maybe(type[input])
+        if input.is_a? Kleisli::Maybe
+          Maybe(type[input.value])
+        else
+          Maybe(type[input])
+        end
       end
       alias_method :[], :call
     end
--- a/spec/dry/types/type_spec.rb
+++ b/spec/dry/types/type_spec.rb
@@ -129,5 +129,19 @@ RSpec.describe Dry::Types::Type do
       expect(value).to be_instance_of(Kleisli::Maybe::Some)
       expect(value.fmap(&:downcase).fmap(&:upcase).value).to eql('SOMETHING')
     end
+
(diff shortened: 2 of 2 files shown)


marshall-lee:
I wonder is it a way to go or not?

solnic:
I'm not sure to be honest. I'll investigate this later next week.

marshall-lee:
@solnic Anyway, don't you think that `Some("SomeString")` should typecheck with `Types::Maybe::Strict::String`?

marshall-lee:
Now I'm not sure too. Constructor should not be called on objects constructed by itself (if it's not a recursive data type, of course).

solnic:
I'll be refining behavior of `Type` objects a bit for the next release, so *maybe* this issue will go away as a side-effect of this refactoring. At least that's what I'm hoping for hah!

PR #30, opened 2016-02-08
Rename to trivial_constructor.

marshall-lee:
It doesn't clear from name what the method does and it's easily confused with `#constructor` instance method.
Ideally, it should be renamed to _identity_, but as it's used as a constructor it is still a constructor. But trivial.


solnic:
In my planned refactoring this constructor will be removed completely. We'll have types with and w/o constructors.

PR #32, opened 2016-02-10
Removing duplication of logic

jeremyf:
Prior to this commit, the code for Dry::Types::Schema.attributes was
doing the following (distilled into a more concrete example):

```ruby
prev_schema = schema || {}

def schema
  @schema || {}
end
```

When expanded that looked something like the following:

```ruby
prev_schema = (@schema || {}) || {}
```

This change is meant to reduce the above to:

```ruby
prev_schema = @schema || {}
```

### DIFF
--- a/lib/dry/types/struct.rb
+++ b/lib/dry/types/struct.rb
@@ -15,7 +15,7 @@ module Dry
       end
 
       def self.attributes(new_schema)
-        prev_schema = schema || {}
+        prev_schema = schema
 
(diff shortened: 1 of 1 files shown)


jeremyf:
Ping @solnic

solnic:
I remember about this. I'll revisit when I'm done with refactoring. Thank you.

solnic:
thanks @jeremyf this makes sense :)

PR #57, opened 2016-03-27
Dry::Types::Struct should be allowed to be blank

joevandyk:


joevandyk:
closing in favor of #58

Issue #15, opened 2016-01-11
Behavior when a struct type does not specify any attributes


backus:
While integrating dry-data into a small project of mine I happened to try to initialize a dry struct before specifying any attributes

```ruby
>> require 'dry/data'
=> true
>>
>> class User < Dry::Data::Struct
 | end
=> nil
>>
>> params = {}
=> {}
>>
>> User.new(params)
NoMethodError: undefined method `[]' for nil:NilClass
from /Users/johnbackus/.rvm/gems/ruby-2.3.0/gems/dry-data-0.4.2/lib/dry/data/struct.rb:34:in `new'
```

Personally, I think this should just initialize an object with no members. I'm not sure what you intend the behavior will be in this case though. It could also throw a custom error saying you can't create structs with zero attributes.

I'd be happy to contribute a fix once I know how this case should behave.

solnic:
Thanks for the report. I think I'd vote for raising an error when no schema is defined for attributes since it's basically a broken class definition. But I gotta think it through

solnic:
I'm gonna close this, handling this gracefully would require additional code that I don't think is worth the effort. An empty struct class is broken after all, it's not gonna work in any sensible way, and raising a meaningful error would require performing a check in `.new` so it would happen every time you instantiate a struct.

PR #58, opened 2016-03-27
Dry::Types::Struct should be allowed to be blank

joevandyk:


spec/dry/types/struct_spec.rb
```diff
@@ -91,6 +91,11 @@ class Test::Foo < Dry::Types::Struct
         end
       }.to raise_error(ArgumentError)
     end
+
+    it 'works on blank structs' do
+      class Test::Foo < Dry::Types::Struct; end
+      expect(Test::Foo.new).to be_kind_of Test::Foo
```
solnic: could you add a check that `#to_h` returns an empty hash too?

joevandyk: updated, also moved spec to better location

solnic:
Refs #15

PR #61, opened 2016-04-01
Add JSON::* coercions

coop:


solnic:
Would it be possible to share some coercions between `Form` and `JSON`?

PR #64, opened 2016-04-03
Improve test coverage for `Struct`, kill dead code, and tighten interface

backus:
I'm using mutant to reveal which parts of `Dry::Types::Struct` are not covered by tests. So far I have addressed the following mutations with (somewhat) opinionated changes.

**1. `attr_reader` mutation**

Passing a key to `attr_reader` twice shouldn't matter in terms of behavior
or performance so I figured it was safe to make this change:

```diff
@@ -1,9 +1,9 @@
 def self.attributes(new_schema)
   prev_schema = schema
   @schema = prev_schema.merge(new_schema)
   @constructor = Types["coercible.hash"].public_send(constructor_type, schema)
-  attr_reader(*(new_schema.keys - prev_schema.keys))
+  attr_reader(*new_schema.keys)
   equalizer.instance_variable_get("@keys").concat(schema.keys).uniq!
   self
 end
```

**2. non-unique `@keys` on `Struct`'s equalizer**

This mutation is more tricky. The current release of dry-types lets me do this:

```ruby
class Foo < Dry::Types::Struct
  attribute :foo, 'strict.string'
  attribute :foo, 'strict.string'
end
```

This behavior should be rejected. It makes our code simpler if we don't have to
think about cleaning up after a sloppy user who repeats attribute definitions.

This also benefits the user because it makes it impossible to either

1. add dead code by repeating the same type definition
2. overwrite a type definition by accident

Therefore I've added a check to `Struct.attributes` which will throw an error if the user redefines an attribute.

**3. Returning `self` from `attributes`**

I added a test for this behavior since `Struct.attributes` and `Struct.attribute`
both are definitely public API and I'm sure the returning `self` behavior is
intentional.

**4. Calling `.new()` on an empty struct subclass**

I've added a test guaranteeing the behavior of empty structs being initialized with a hash if `.new` is called without arguments. This is the current behavior so it is probably important this behavior doesn't regress. Alternatively I think it would also be appropriate to make it impossible to initialize a struct with zero arguments

### DIFF
--- a/lib/dry/types/errors.rb
+++ b/lib/dry/types/errors.rb
@@ -18,6 +18,12 @@ module Dry
 
     StructError = Class.new(TypeError)
 
+    class RepeatedAttributeError < ArgumentError
+      def initialize(key)
+        super("Attribute :#{key} has already been defined")
+      end
+    end
+
     ConstraintError = Class.new(TypeError) do
       attr_reader :result
 
--- a/lib/dry/types/struct.rb
+++ b/lib/dry/types/struct.rb
@@ -63,8 +63,8 @@ module Dry
         else
           super(constructor[attributes])
         end
-      rescue SchemaError, SchemaKeyError => e
-        raise StructError, "[#{self}.new] #{e.message}"
+      rescue SchemaError, SchemaKeyError => error
+        raise StructError, "[#{self}.new] #{error}"
       end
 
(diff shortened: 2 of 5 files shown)


solnic:
These are very good improvements and fixes. Thank you for the comprehensive description and reasoning <3

Issue #47, opened 2016-03-24
Add support for Symbol primitive


solnic:
Would be nice to add support for:

- `Nominal::Symbol`
- `Strict::Symbol`
- `Coercible::Symbol`

joevandyk:
i'll take a stab at this tonight

solnic:
This was already done last year, by me, oh well.

morr:
@solnic `Symbol` is put in `NON_COERCIBLE` in this commit https://github.com/dry-rb/dry-types/commit/761e75902a5c063116572c6fe9e25b3d2660d430
Can you make it coercible again?

morr:
i found workaround but it looks very ugly
```ruby
Types::Strict::Symbol
      .constructor(-> (val) { val.to_sym })
      .enum(:a, :b, :c)
```

solnic:
@morr `Coercible` uses coercion methods from `Kernel` and there's no `Kernel.Symbol` so we won't have coercible symbol. We could add coercible symbol to other types, like form or json.

solnic:
One more thing:

> i found workaround but it looks very ugly

What's ugly about it? It's not a workaround, it's *the* way of using this library :)

morr:
Because of the way of using it for strings :)

```ruby
Types::Coercible::String('a', 'b', 'c')
```

vs

```ruby
Types::Strict::Symbol
  .constructor(-> (val) { val.to_sym })
  .enum(:a, :b, :c)
```

solnic:
@morr you can just do this:

``` ruby
Types::Strict::Symbol.constructor(&:to_sym).enum(:a, :b, :c)
```

We can also think about adding a new type category for `to_*` coercions, so we have `Coerible` which uses `Kernel` coercion functions, and we'd have another one that uses `to_s` `to_sym` `to_i` and so on.

georgemillo:
[...]

Issue #46, opened 2016-03-24
Rename `Coercible::*` to `Kernel::*` for clarity


solnic:
I believe the `Coercible` types may cause confusion as people may expect complex coercion logic under this category whereas it only provides `Kernel` type constructor methods (like `Kernel.Integer` etc.).

mrbongiolo:
It makes sense. Later on you can actually use the `Coercible` namespace for more complex coercions, if needed or if someone just wants to extend Dry::Types.

daniels:
As one who didn't realize that the coercions in this namespace were only expected to utilize the `Kernel` methods I'm still note entirely sure about the name change. It says more about the implementation but less about what's special about these types. I think I get more information from `Types::Coercible::String` than from `Types::Kernel::String`.

joevandyk:
What's the reasoning for only doing Kernel coercions? Simplicity?

solnic:
@joevandyk intentional usage. ie kernel coercions can be useful in low-level library code.

`Coercible::*` is not going to be very common but there are use cases for it. I'm planning to add more categories, like `JSON::*` (which will probably share a lot with `Form::*`) and probably more once we discover new use cases.

What I want to avoid is "generic coercions" that we have in libs like virtus and its clones. This approach has failed and I don't want to repeat the same mistake.

Please also see my [comment](https://github.com/dry-rb/dry-types/issues/44#issuecomment-201214855) in the related #44

PR #66, opened 2016-04-03
Add coercible.symbol type

backus:
I believe this addresses #47. The issue also mentions `strict.symbol` but it looks like @solnic added that a while ago https://github.com/dry-rb/dry-types/commit/761e75902a5c063116572c6fe9e25b3d2660d430.

### DIFF
--- a/lib/dry/types/coercions/symbol.rb
+++ b/lib/dry/types/coercions/symbol.rb
@@ -1,0 +1,20 @@
+require 'date'
+require 'bigdecimal'
+require 'bigdecimal/util'
+require 'time'
+
+module Dry
+  module Types
+    module Coercions
+      module Symbol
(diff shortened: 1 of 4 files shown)


solnic:
For `Coercible` we only support types that have corresponding Kernel coercion methods, so it's "plain ruby". There's no `Symbol(input)` kernel method, so we can't support it. There was a recent discussion in #46 whether we should rename `Coercible::*` to something else to indicate what this category is for.

backus:
Ah alright. Well what did issue #47 mean then? I'm confused

solnic:
Well, it was me being tired not remembering that I already added `Symbol` :( Sorry for confusion

backus:
Heh no problem.

PR #68, opened 2016-04-03
Improve test coverage for `Result`, `Definition`, and `JSON.to_decimal`

backus:
- 100% mutation coverage on `JSON.to_decimal`
- Brings mutation coverage of Result* up to 100% from 58.33%
- Specifies behavior of Definition#try

Result::Success and Result::Failure seem simple enough to assume that
they are implementation details so I tested the behavior through
Definition#try

### DIFF
--- a/spec/dry/types/definition_spec.rb
+++ b/spec/dry/types/definition_spec.rb
@@ -16,33 +16,37 @@ RSpec.describe Dry::Types::Definition do
   end
 
   describe '#try' do
-    subject(:try) { type.try(value) }
-
-    let(:success) { try.success?    }
-    let(:failure) { try.failure?    }
-    let(:input)   { try.input       }
+    let(:result) { type.try(value) }
 
     context 'when given valid input' do
       let(:value) { 'foo' }
 
-      specify { expect(success).to be(true)  }
-      specify { expect(failure).to be(false) }
-      specify { expect(input).to be(value)   }
+      it 'returns a success' do
+        expect(result).to be_success
+      end
+
+      it 'provides the original input' do
+        expect(result.input).to be(value)
+      end
     end
 
     context 'when given invalid input' do
       let(:value) { :foo }
 
-      specify { expect(success).to be(false) }
-      specify { expect(failure).to be(true)  }
-      specify { expect(input).to be(value)   }
+      it 'returns a failure' do
+        expect(result).to be_failure
+      end
+
+      it 'provides the original input' do
+        expect(result.input).to be(value)
+      end
 
       it "provides an error message" do
-        expect(try.error).to eql(':foo must be an instance of String')
+        expect(result.error).to eql(':foo must be an instance of String')
       end
 
       it "yields failure when given a block" do
-        expect { |probe| type.try(value, &probe) }.to yield_with_args(try)
+        expect { |probe| type.try(value, &probe) }.to yield_with_args(result)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/dry/types/definition_spec.rb
```diff
@@ -16,33 +16,37 @@
...
+    let(:result) { type.try(value) }
 
     context 'when given valid input' do
       let(:value) { 'foo' }
 
-      specify { expect(success).to be(true)  }
-      specify { expect(failure).to be(false) }
-      specify { expect(input).to be(value)   }
+      it 'returns a success' do
+        expect(result).to be_success
```
backus: I opted to not use `be_success` before since it asserts truthyness and not that `success?` returns `true`

solnic:
Thanks. I refactored the spec to match with the spec style we have. It is preferred when `subject` is used exclusively for the object under test, and avoid using additional let statements to make a spec more concise. People actually have trouble following specs when things are done in a smart way. Oh and please don't do the right-side alignment of brackets, curly braces or parenthesis, it's such a weird thing to do which adds additional maintenance cost without any actual benefit other than "it kinda looks nice".

I guess I should document it somewhere :(

backus:
Sure thing.

> I guess I should document it somewhere :(

Even if it isn't part of the build system, the most helpful thing for me would be rubocop settings included in the project.

spec/dry/types/definition_spec.rb
solnic: It's unfortunate, I prefer when predicate methods return either `true` or `false`, but having said that, I don't see any good reason to be very conservative about it in tests. Mut cov remained 100% anyway :)

Specs should be focusing on behavior, and here behavior is that `if success?; puts 'ok'; else puts 'not ok'; end` works as expected, not that the return value is `true` and not `1`.

solnic:
Is it possible to configure rubocop to check rspec style? That would be interesting.

backus:
> Is it possible to configure rubocop to check rspec style? That would be interesting.

Yes. See: [rubocop-rspec](https://github.com/nevir/rubocop-rspec)

I'm also writing my own and it is fairly easy. For example: [`Devtools/NamedSubject`](https://github.com/backus/rubocop-devtools/blob/master/lib/rubocop/cop/devtools/named_subject.rb#L4-L34)

solnic:
Can we configure that in CodeClimate?

backus:
> Can we configure that in CodeClimate?

No idea. I've never tried it

solnic:
Ah OK, so basically CC now supports configuring rubocop using your own settings, so theoretically we should be able to provide them and things should Just Work™. It's simple to look at a branch in CC and see if things gotten better or worse, or stayed the same. Having this setup on CC side lowers maintenance cost that's why I moved it there.

backus:
Cool. Well I'd be interested if code climate Just Works with custom cops. 

> Having this setup on CC side lowers maintenance cost that's why I moved it there.

Yup I understand (I'm aware of where you stand on the whole [rom-rb/devtools](https://github.com/mbj/devtools) thing) so I didn't mean to seem like I'm pushing for that. I meant more of like a rubocop file in the project that I can use to check my code.

backus:
> I'm aware of where you stand on the whole rom-rb/devtools thing

citation 😆  https://www.youtube.com/v/LiRYqor9ogs?start=2001&end=2016&autoplay=true

solnic:
Haha, glad we did this session with Markus, at least I don't need to repeat myself. FWIW I haven't changed my mind :)

PR #69, opened 2016-04-04
Inherit the constructor_type on Struct

fbernier:
Previous behavior was

```ruby
  constructor_type(:schema)
end

class Admin < User; end;

p User.constructor_type
=> :schema
p Admin.constructor_type
=> :strict
```

This PR makes sure `Admin`'s constructor type would be `:schema` here.


solnic:
Thank you

PR #71, opened 2016-04-05
Make safe schema intolerant to nils for strict types

flash-gordon:
I believe with Optional type (that actually is just an instance of Sum type) we can bring more strictness to Safe/Symbolized schemas. The constructor will set nils automatically if you define type as optional. That seems to me more consistent than having an object with empty values for strict types:
```ruby
gordon@mac ~/dev/dry-types [master] $ ./bin/console
2.3.0 :001 > class Foo < Dry::Types::Struct
2.3.0 :002?>   constructor_type :schema
2.3.0 :003?>   attribute :required, Dry::Types['strict.int']
2.3.0 :004?> end
 => Foo
2.3.0 :005 > Foo.new({})
 => #<Foo required=nil>
2.3.0 :006 >
```

### DIFF
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -25,7 +25,7 @@ module Dry
         def resolve_missing_value(result, key, type)
           if type.is_a?(Default)
             result[key] = type.evaluate
-          elsif type.is_a?(Maybe)
+          else
             result[key] = type[nil]
           end
         end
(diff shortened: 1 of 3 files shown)


flash-gordon:
I tried to implement timestamp attributes with brand new shiny schema for rom-relation from master branch. And that's what I did:
```ruby
module Types
  CreatedAt = ::Types::Strict::Time.default(-> { ::Time.current })
  UpdatedAt = CreatedAt.constructor(-> _ { ::Time.current })
end
```
As you can see my idea was to link `updated_at` attribute with a constructor type. But I failed because atm constructor types are just skipped by `Schema::Safe` constructor. So I changed the order of calling `#default` and `#constructor` like this
```ruby
module Types
  CreatedAt = ::Types::Strict::Time.default(-> { ::Time.current })
  UpdatedAt = ::Types::Strict::Time.constructor(-> _ { ::Time.current }).default(-> { ::Time.current })
end
```
and not it works. It was a little bit tricky :)

PR #75, opened 2016-04-10
Fix compiler crash on empty hash schema, 74

### LINKED ISSUE #74: Compiler crashes when hash type has an empty schema
`[:type, ["hash", [:schema, []]]` causes the compiler to crash, it should be allowed to create a hash with empty schema.

refs dry-rb/dry-validation#101

lasseebert:
Fixes #74 

As I'm new to dry-types, I have very little overview if this could have negative impact elsewhere.


solnic:
Thank you for spotting this and fixing :)

lasseebert:
Thanks for the good explanation in the issue :)

Issue #78, opened 2016-04-19
conflict with active_support :try


andreaseger:
The internal use of the method `try` in combination with method_missing runs into an error as soon as active_support with it's `try` core_ext is required.

```
require 'bundler/inline'
gemfile true do
  gem 'dry-types'
  gem 'activesupport', require: false
end

include Dry::Types.module

CustomType = Strict::String.enum("val1", "val2")
CustomStringArray = Strict::Array.member(CustomType)

p CustomStringArray.call(["val1"])

require "active_support/core_ext/object/try"

p CustomStringArray.call(["val1"])
```

I'm not sure the method_missing cascade is used similar in other scenarios.

Given how common active_support is even outside rails applications this seems like a issue that should be fixed. Without to much insight into the code renaming the try method to something unique or changing how the method_missing cascade is used here might be an option so solve this.

tak1n:
The problem here is that ActiveSupport monkey patches `Object` which `Dry::Types::Enum` implicitly inherits from.

```
From: /home/benny/Dev/ruby/dry-rb/bugs/.gem/ruby/2.3.0/gems/dry-types-0.7.1/lib/dry/types/array/member.rb @ line 15 Dry::Types::Array::Member#call:

    12: def call(input, meth = :call)
    13:   input.map { |el|
    14:     result = member.__send__(meth, el)
 => 15:     binding.pry
    16:     result
    17:   }
    18: end

[1] pry(#<Dry::Types::Array::Member>)> meth
=> :try
[2] pry(#<Dry::Types::Array::Member>)> member
=> #<Dry::Types::Enum:0x0055df70d72fa8
 @mapping={0=>"val1", 1=>"val2"},
 @options={:values=>["val1", "val2"]},
 @type=
  #<Dry::Types::Constrained:0x0055df70d73070
   @options=
    {:rule=>
      #<Dry::Logic::Rule::Conjunction left=#<Dry::Logic::Rule::Value predicate=#<Dry::Logic::Predicate id=:type? args=[String]> options={}> right=#<Dry::Logic::Rule::Value predicate=#<Dry::Logic::Predicate id=:inclusion? args=[["val1", "val2"]]> options={}>>},
   @rule=
[...]

PR #79, opened 2016-04-20
make Dry::Types::Enum inherit from BasicObject

andreaseger:
as further discussion base to fix the issues described in #78 

As @timriley suggested I quickly tried using `BasicObject` for the other 5 classes which mixin `Decorator`.
In all cases the specs start to fail in - lets say interesting - manner. But I didn't dig further into it.

### DIFF
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -2,7 +2,7 @@ require 'dry/types/decorator'
 
 module Dry
   module Types
-    class Enum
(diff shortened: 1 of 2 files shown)


solnic:
I need to think how to solve this. BasicObject is an option. I need to investigate what makes sense.

solnic:
I just defined `try` explicitly in the decorator so #78 is now fixed, gonna close this one

PR #82, opened 2016-04-26
Capitalised versions of true and false values

nil0bject:
Please can I have this change? It's very convenient to have these extra values.
I can refactor, as the values are just repeating and being transformed in some way, but I would need guidance. The codebase is a bit beyond me.


solnic:
Yes you can, thank you <3

nil0bject:
Thank you!

PR #84, opened 2016-04-30
Let Form::Date[nil] return nil

### LINKED ISSUE #83: (Types::Form::Date | Types::Strict::Nil) not working for nil?
Using gem version `0.7.1`.

I might be misunderstanding `|`, but I expected it to work as a fallback if the left coercion fails.

Code
---

```ruby
(Types::Form::Date | Types::Strict::Nil)[nil]
```

Result I get
---

```
Dry::Types::ConstraintError: nil violates constraints (#<struct Dry::Types::Result::Failure input=nil, error="no implicit conversion of nil into String">)
	from dry-types-0.7.1/lib/dry/types/sum.rb:31:in `block in call'
	from dry-types-0.7.1/lib/dry/types/sum.rb:44:in `try'
	from dry-types-0.7.1/lib/dry/types/sum.rb:30:in `call'
```

Result I expected
---

```ruby
nil
```

`(Types::Form::Decimal | Types::Strict::Nil)[nil]` works btw, so it seems to be specific to the `Form::Date` type.

timhabermaas:
`Date.parse(nil)` raises a `TypeError`, so handle that case explicitly.
I could've rescued from `TypeError`, but that would allow for
`Form::Date[2]` to return `2`. That's probably not how form coercions should
behave.

This fixes #83.

### DIFF
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -8,6 +8,8 @@ module Dry
       end
 
       def to_date(input)
+        return nil if input.nil?
+
         Date.parse(input)
       rescue ArgumentError
         input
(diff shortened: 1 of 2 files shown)


solnic:
Form types are only meant to work with string input so this behavior is not needed.

PR #85, opened 2016-05-03
Replace kleisli with dry-monads (closes #80)

### LINKED ISSUE #80: Replace kleisli with dry-monads
It looks like we have some issues with kleisli gem. First of all the author didn't release the new gem version that would include much welcome removal of obsolete blank_slate hack. When I tried to add dry-types to my existing rails project this first thing I saw was StackOverflowError. It if very unfriendly error tbh :) After some time of debugging I figured out that the problem was caused by [other gem](https://github.com/kschiess/parslet/pull/154) that uses blankslate (without reason, actually). I absolutely don't blame the author at all, it is OSS so he can do whatever he wants. 

The second issue is that kleisli adds some pollution to the global environment. E.g. it adds `#Maybe` and `#Right` methods to object so the become a private members of almost any object in runtime. I think it is against our rules :) Also I Maybe/Either methods a lot and found a tricky problem with using them inside a namespaced blocks of dry-container because [NamespaceDSL](https://github.com/dry-rb/dry-container/blob/master/lib/dry/container/namespace_dsl.rb) uses SimpleDelegator that inherits from BasicObject and _does not_ delegate to private methods so I got really unexpected NoMethodError.

Third is that I think we do not need most of kleisli classes. I doubt we need more than Maybe + Either. And from the other side we can add some other stuff to a potential monad/fp dependency.

Finally I think the API of kleisli is quite good but not perfect. Twisting it a bit would be really nice.

I propose to create a minor dry-monad/dry-fp dependency so we can solve all that issues. I quickly made a couple of files (untested :)) that can become a new gem:
Maybe: https://gist.github.com/flash-gordon/a6e6c0956b802d387e2fe81d9b7530b7
Either: https://gist.github.com/flash-gordon/446d38a1c7295e1b9ffcbe1302741522

Any thoughts?

flash-gordon:
After releasing dry-monads we are now able to replace kleisli gem.

### DIFF
--- a/dry-types.gemspec
+++ b/dry-types.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'dry-configurable', '~> 0.1'
   spec.add_runtime_dependency 'dry-logic', '~> 0.2', '>= 0.2.0'
   spec.add_runtime_dependency 'inflecto', '~> 0.0.0', '>= 0.0.2'
-  spec.add_runtime_dependency 'kleisli', '~> 0.2'
+  spec.add_runtime_dependency 'dry-monads', '>= 0.0.1'
 
   spec.add_development_dependency "bundler", "~> 1.6"
   spec.add_development_dependency "rake", "~> 11.0"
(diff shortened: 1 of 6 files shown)


flash-gordon:
I think it can be merge into master before 0.8 is released. Does anyone have concerns about it?
/cc @timriley @AMHOL

AMHOL:
Looks good to me @flash-gordon

timriley:
Looks good, thanks!

PR #96, opened 2016-06-03
Coerce nil values in form to blank arrays and hashes

kirs:
Basically the same thing as #88, but for nil values.

review @timriley @flash-gordon

### DIFF
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -30,6 +30,10 @@ module Dry
       def empty_str?(value)
         EMPTY_STRING.eql?(value)
       end
+
+      def empty?(value)
+        value.nil? || empty_str?(value)
+      end
     end
   end
 end
--- a/lib/dry/types/coercions/form.rb
+++ b/lib/dry/types/coercions/form.rb
@@ -50,11 +50,11 @@ module Dry
         end
 
         def self.to_ary(input)
-          empty_str?(input) ? [] : input
+          empty?(input) ? [] : input
         end
 
         def self.to_hash(input)
-          empty_str?(input) ? {} : input
+          empty?(input) ? {} : input
         end
       end
     end
--- a/spec/dry/types/types/form_spec.rb
+++ b/spec/dry/types/types/form_spec.rb
@@ -192,6 +192,11 @@ RSpec.describe Dry::Types::Definition do
       expect(type[input]).to eql([])
     end
 
+    it 'coerces nil into an empty array' do
+      input = nil
+      expect(type[input]).to eql([])
+    end
+
     it 'returns original value when it is not an array' do
       foo = 'foo'
       expect(type[foo]).to be(foo)
@@ -208,6 +213,11 @@ RSpec.describe Dry::Types::Definition do
 
     it 'coerces an empty string into an empty hash' do
       input = ''
+      expect(type[input]).to eql({})
+    end
+
+    it 'coerces nil into an empty hash' do
+      input = nil
       expect(type[input]).to eql({})
     end
 


timriley:
Thanks for looking into this! However, I don't think we would want this behaviour in the form types, since Rack-processed form posts can never have `nil` values. Adding this behaviour would mean that form types would no longer be focused on processing form output only.

I'm going to close this for now, but I'd still be interested to hear what scenario led you to require this kind of behaviour. Perhaps we can find some other ways to help!

kirs:
Thanks for a detailed reply. At first I was going to answer with "gotcha" but today I could find the use case when request parameters come already coerced.

Steps to reproduce:

1. Create a simple Rails controller that responds with the same input it received:

```ruby
  def create
    render plain: request.parameters.inspect
  end
```

2. Send a JSON request to this controller: 
```
curl -X POST -H "Content-Type: application/json" -d '{"player": {"name": "test","room_id": 0,"skin_id": 1,"head_id": 2,"torso_id": 3,"legs_id": 4,"x_position": 5,"y_position": 6,"direction": true,"action": "","gender": "f"}}' localhost:3000/posts
```

3. Rails coerces the JSON input:

```
Processing by PostsController#create as */*
  Parameters: {"player"=>{"name"=>"test", "room_id"=>0, "skin_id"=>1, "head_id"=>2, "torso_id"=>3, "legs_id"=>4, "x_position"=>5, "y_position"=>6, "direction"=>true, "action"=>"", "gender"=>"f"}, "post"=>{}}
```

Here is the part of Rails that does coercion on JSON requests: https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/parameters.rb#L8

So the input already comes typecasted and we cannot simply pass it to `Dry::Validation.Form`.
I understand that obviously `dry-rb` maintainers wouldn't want to see hacks especially for Rails.

For now I'm going to solve this issue with a custom `input_processor`, but I'd like to discuss how Dry::Validation could help to solve this case (and improve the adoption in Rails apps that use JSON API features).

@solnic @timriley looking forward to know your opinions.

kirs:
This is probably concern of `Dry::Validation.JSON` but in my case I use the same schema class for both form and JSON requests.

timriley:
Thanks @kirs for the extra info here!

I'm wondering: if your input data is already coerced thanks to Rails, is there any reason for you to use a form or JSON schema in the first place? Could you just work with an ordinary dry-validation schema?

kirs:
> is there any reason for you to use a form or JSON schema in the first place? 

It is, because not all types are coerced (e.g. date / time)
In my case, I wanted to use the same schema class for **both** form and JSON requests.

solnic:
We should probably figure out how to handle such cases. I suspect this won't be a rails-specific issue anyway. I need to think about this and get back to you soon.

solnic:
I should mention that it is possible to create a base schema and then create 2 separate schemas for forms and json from the base one, they would configure their coercions and inherit all rules from the base one. Is this something that sounds like a good solution @kirs?

kirs:
Thanks, that sounds exactly what I'm looking for. Could you share some snippets for that?

solnic:
@kirs will do once I actually make this work :D

flash-gordon:
This is kind of outdated. Nevertheless, you can build two schemas on top of common definition:

```ruby
CommonSchema = Dry::Validation.Schema(build: false) do
  # ...
end

FormSchema = Dry::Validation.Schema(CommonSchema) { 
  configure { config.input_processor = :form } 
}

JSONSchema = Dry::Validation.Schema(CommonSchema) { 
  configure { config.input_processor = :json } 
}
```

PR #98, opened 2016-06-16
Make Struct#to_hash go through arrays

nepalez:
`Struct#to_hash` hashifies stuctures deeper (going inside arrays).

Suppose a struct has array attributes with struct members:
```
  attribute :name, Types::Coercible::String
end

class Parent < User
  attribute :children, Types::Array.member('user')
end

attributes = { name: 'Joe', children: [{ name: 'Jane' }] }
parent = Parent.new(attributes)
```
Old behaviour:
```
parent.to_hash # => { name: 'Joe', children: [#<User name="Jane">] }
```
New behaviour:
```
parent.to_hash # => { name: 'Joe', children: [{ name: 'Jane' }] }
```

### DIFF
--- a/lib/dry/types/struct.rb
+++ b/lib/dry/types/struct.rb
@@ -1,3 +1,5 @@
+require_relative 'hashify'
+
 module Dry
   module Types
     class Struct
@@ -87,10 +89,9 @@ module Dry
       end
 
       def to_hash
-        self.class.schema.keys.each_with_object({}) { |key, result|
-          value = self[key]
-          result[key] = value.respond_to?(:to_hash) ? value.to_hash : value
-        }
+        self.class.schema.keys.each_with_object({}) do |key, result|
+          result[key] = Hashify[self[key]]
+        end
       end
       alias_method :to_h, :to_hash
     end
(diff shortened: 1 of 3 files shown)


lib/dry/types/struct.rb
```diff
@@ -87,10 +89,9 @@ def [](name)
       end
 
       def to_hash
-        self.class.schema.keys.each_with_object({}) { |key, result|
-          value = self[key]
-          result[key] = value.respond_to?(:to_hash) ? value.to_hash : value
-        }
+        self.class.schema.keys.each_with_object({}) do |key, result|
+          result[key] = Hashify[self[key]]
```
solnic: it's an overkill to instantiate Hashify object here, could we simply provide this functionality in this method instead?

nepalez: D'oh! You're right about instantiations. I should have implemented hashify as a utility module (and I will).

I couldn't drop external hashifier and implement the functionality in the method because we need going through many nested arrays (or sets etc.).

Their items do not necessarily implement `#to_hash` by themselves, so we need to check `to_hash` and `map` at any layer by some external visitor/hashifier.

PR #99, opened 2016-06-18
Deep freeze Dry::Types::Value objects

### LINKED ISSUE #65: Freeze attributes belonging to Dry::Types::Value objects?
Initializing a struct or value object with an existing instance just returns that object:

```ruby
require 'dry/types'

class Name < Dry::Types::Value
  attribute :first, 'strict.string'
  attribute :last,  'strict.string'
end

name1 = Name.new(first: 'john', last: 'doe') # => #<Name first="john" last="doe">
name2 = Name.new(name1)                      # => #<Name first="john" last="doe">

name1.equal?(name2) # => true

name1.first.reverse!

name1.first # => "nhoj"
name2.first # => "nhoj"
```

I would say that `dry-types` should *at least* freeze each attribute's value. I realize that a `Dry::Types::Value` instance itself is frozen but its members should probably also be frozen.

backus:
Closes #65


backus:
@solnic this is ready for review

solnic:
Thanks :)

PR #101, opened 2016-06-22
Make Struct work as a type definition

flash-gordon:
This add an interface of a type definition to Struct and Value and makes them work with `.|`, `.default`, `.enum`, and `.optional` methods for building new types. This is became a frequently requested feature and as far as I can tell nothing stops us from supporting it. A noticeable change is that Struct/Value subclasses are now registered in the container as is, we don't need to wrap them with a type definition anymore.
Any thoughts? /cc @solnic @AMHOL @timriley

### DIFF
--- a/spec/shared/struct.rb
+++ b/spec/shared/struct.rb
@@ -55,7 +55,7 @@ RSpec.shared_examples_for Dry::Types::Struct do
 
   context 'class interface' do
     describe '.|' do
-      let(:sum_type) { type_or_nil = type | Dry::Types['strict.nil'] }
+      let(:sum_type) { type | Dry::Types['strict.nil'] }
 
       it 'returns Sum type' do
         expect(sum_type).to be_instance_of(Dry::Types::Sum)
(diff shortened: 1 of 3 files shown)


solnic:
Yeah I think this makes a lot of sense, I also just realized that if we don't auto-register struct/value then the whole issue with rails code reloading should be gone...

spec/shared/struct.rb
```diff
@@ -76,16 +44,68 @@
...
       expect(user_1.inspect).to eql(
-        %Q(#<#{type.primitive} name="Jane" age=21 address=#<Test::Address city="NYC" zipcode="123"> root=true>)
+        %Q(#<#{type} name="Jane" age=21 address=#<Test::Address city="NYC" zipcode="123"> root=true>)
       )
     end
   end
+
+  context 'class interface' do
+    describe '.|' do
+      let(:sum_type) { type_or_nil = type | Dry::Types['strict.nil'] }
```
solnic: why do we need `tyoe_or_nil` here?

flash-gordon: oh, of course we don't need it

flash-gordon:
@solnic I removed the variable, anything else? I'm going merge it

solnic:
Merged! woohaa, thanks man. We gotta ask people on rails if this resolves the reloading problem :)

PR #118, opened 2016-07-29
Improve dry/types_spec.rb

abinoam:
This PR tries to make a clear distinction of a constructor that calls the default _new_ methods and a constructor that calls a custom method (ex: build) differentiating them with a plain and an upcased result.


solnic:
That's a nice improvement. Thank you :)

PR #119, opened 2016-08-05
Improve form boolean casting

kirs:
* typecast `T` and `F` to boolean values (before we only accepted `t` and `f`)
* stringify boolean input values before passing them to map

The second thing may sound like it's not the case for `Form`, but here is my use case:

`Dry::Validation.Form` receives input from JSON parsed by Rails.
The clients sends the following JSON:
```json
{ "accept_the_rules": 0 }
```

which will be parsed by Rails JSON parser to

```ruby
{ "accept_the_rules" => 0 }
```

And `Dry::Validation.Form` will receive the integer value that it will try to typecast to boolean.

@solnic what do you think?

### DIFF
--- a/lib/dry/types/coercions/form.rb
+++ b/lib/dry/types/coercions/form.rb
@@ -12,11 +12,11 @@ module Dry
         extend Coercions
 
         def self.to_true(input)
-          BOOLEAN_MAP.fetch(input, input)
+          BOOLEAN_MAP.fetch(input.to_s, input)
         end
 
         def self.to_false(input)
-          BOOLEAN_MAP.fetch(input, input)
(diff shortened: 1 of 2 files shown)


solnic:
Thanks, this makes sense. I should somehow split json/form logic in a nicer way though, but it can stay like that for now.

kirs:
> I should somehow split json/form logic in a nicer way though

I think what we should do is to make form types tolerant to all kinds of input, for instance for integers both `1` and `"1"` and so on.

In this case we could use `Schema` for strict input and `Form` for non-strict.

solnic:
@kirs in #138 we're gonna make date/time-related Form types tolerant to all kinds of input, this will make it work consistently with other Form types

PR #121, opened 2016-08-10
Adds '.include?' to enum.

d-Pixie:
Mostly to support doing this:

``` ruby
CountryCodes = Types::Strict::String.enum( 'AF','AX',...,'ZM','ZW' )
CountryCode = Types::Strict::String.constrained(included_in: CountryCodes).constructor{|v| v.to_s.upcase[0...2]}
```

There might be a better way of doing this (right now I'm using `included_in: CountryCodes.values`) if so we could add that to the docs.

I originally tried doing it like this:

``` ruby
CountryCodes = Types::Strict::String.enum( 'AF','AX',...,'ZM','ZW' )
CountryCode = CountryCodes.constructor{|v| v.to_s.upcase[0...2]}
```

But that did not seem to work ...

### DIFF
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -17,7 +17,7 @@ module Dry
 
       def call(input)
         value =
-          if values.include?(input)
+          if include?(input)
             input
           elsif mapping.key?(input)
             mapping[input]
@@ -26,6 +26,10 @@ module Dry
         type[value || input]
       end
       alias_method :[], :call
+
+      def include?(input)
+        values.include?(input)
+      end
     end
   end
 end


solnic:
You could use `valid?(input)` to get the same result ie:

``` ruby
Dry::Types['int'].enum(1, 4, 6).valid?(2) # false
Dry::Types['int'].enum(1, 4, 6).valid?(4) # true
```

Would that be enough for you?

d-Pixie:
No, since this is a problem when using an enum type i another types
constraint, included, clause. Since that constraint is implemented to use
`.included?` it needs too be added to the enum class so you don't have to
do `constrained( included_in: enum_type.values )`. Adding `.included?` to
the enum class allows us to drop that `.values` tail in the type definition.

lib/dry/types/enum.rb
```diff
@@ -26,6 +26,10 @@ def call(input)
         type[value || input]
       end
       alias_method :[], :call
+
+      def include?(input)
+        values.include?(input)
+      end
     end
```
solnic: Would `alias_method :include?, :valid?` do the trick?

GustavoCaso: I tried and it will do the trick @solnic @d-Pixie.

The `valid?` method is `Decorator` file if you want to have a look.

https://github.com/dry-rb/dry-types/blob/master/lib/dry/types/decorator.rb#L27

flash-gordon:
jeez, I broke github 🤣 
image

PR #124, opened 2016-08-16
Fix the command to run tests in the README

roryokane:
When I tried running `rake spec` without `bundle exec`, I just got this error:

    ~/.rbenv/versions/2.3.1/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- dry/logic/rule_compiler (LoadError)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


solnic:
thanks!

PR #125, opened 2016-08-16
Small refactoring to reduce duplication in core.rb

roryokane:
I was pointed to this problem by [Code Climate dry-rb/dry-types Issues](https://codeclimate.com/github/dry-rb/dry-types/issues), which listed two instances of “similar code” in this file.


solnic:
Thanks, again :)

PR #132, opened 2016-08-24
Port Cache from rom-support

flash-gordon:
This is the first step of the killing rom-support 🔫

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,9 @@ bundler_args: --without benchmarks tools
 script:
   - bundle exec rake spec
 rvm:
-  - 2.0
-  - 2.1
-  - 2.2
-  - 2.3.0
+  - 2.1.10
+  - 2.2.5
+  - 2.3.1
   - rbx-2
   - jruby-9000
   - ruby-head
--- a/lib/dry/types.rb
+++ b/lib/dry/types.rb
@@ -12,12 +12,14 @@ require 'dry/types/version'
 require 'dry/types/container'
 require 'dry/types/definition'
 require 'dry/types/constructor'
+require 'dry/types/cache'
 
 require 'dry/types/errors'
 
 module Dry
   module Types
     extend Dry::Configurable
+    extend Cache
 
     Undefined = Object.new.freeze
 
@@ -53,7 +55,7 @@ module Dry
     end
 
     def self.[](name)
-      type_map.fetch_or_store(name) do
+      cache.fetch_or_store(name) do
         case name
         when String
           result = name.match(TYPE_SPEC_REGEX)
@@ -87,10 +89,6 @@ module Dry
 
     def self.identifier(klass)
       Inflecto.underscore(klass).tr('/', '.')
-    end
-
-    def self.type_map
-      @type_map ||= Concurrent::Map.new
     end
 
     def self.type_keys
--- a/lib/dry/types/cache.rb
+++ b/lib/dry/types/cache.rb
@@ -1,0 +1,27 @@
+require 'concurrent/map'
+
+module Dry
+  module Types
+    module Cache
+      def self.extended(klass)
+        super
+        klass.include(Methods)
+        klass.instance_variable_set(:@__cache__, Concurrent::Map.new)
+      end
+
+      def cache
+        @__cache__
+      end
+
+      def fetch_or_store(*args, &block)
+        cache.fetch_or_store(args.hash, &block)
+      end
+
+      module Methods
+        def fetch_or_store(*args, &block)
+          self.class.fetch_or_store(*args, &block)
+        end
+      end
+    end
+  end
+end
--- a/spec/dry/types/cache_spec.rb
+++ b/spec/dry/types/cache_spec.rb
@@ -1,0 +1,24 @@
+require 'dry/types/cache'
+
+RSpec.describe Dry::Types::Cache do
+  subject(:klass) do
+    Class.new do
+      extend Dry::Types::Cache
+    end
+  end
+
+  describe '#fetch_or_store' do
+    it 'stores and fetches a value' do
+      args = [1, 2, 3]
+      value = 'foo'
+
+      expect(klass.fetch_or_store(*args) { value }).to be(value)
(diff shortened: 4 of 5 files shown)


solnic:
Dis is gut but maybe we could create something like dry-core and start adding core APIs there (this could include fp stuff too, potentially, eventually, maybe). WDYT?

flash-gordon:
@solnic, _potentially_ and _eventually_ are monads, I guess? :)

flash-gordon:
re dry-core, sure, mb we need to build a shortlist for it or sutin. Eg, dry-equalizer can be there

solnic:
Haha, yeah. So equalizer could be there, esp that we use it literally everywhere. Other stuff is:

* Cache
* Options (from rom-support) we may just use dry-initializer for that assuming it's fast enough
* ClassMacros (from rom-support) - this should be renames though, these are just inheritable class instance vars
* Deprecations
* Logger (maybe?)
*

flash-gordon:
@solnic k, just create a repo, grant me push access, I'll take care of the rest ;)

AMHOL:
@solnic @flash-gordon @timriley @fran-worley  I'm not keen on the idea of dry-core, would rather have separate gems for each and keep each concern separate and independently usable, WDYT?

solnic:
Dunno yet. We shall discuss :)

flash-gordon:
@AMHOL it's OK, it's just the same old problem, absolutely nothing new here :) dry-core is some sort of _hypothetical_ standard library for dry-rb gems. This implies certain restrictions on what we can put there and how that should be used, so, probably, this would resolve your concerns.
Those things shouldn't be changed often. For instance, dry-equalizer is pretty much done and no one really knows if it works with MRI 2.3, because the later is missing in travis' config :) I bet the same thing will happen with deprecations and other guys from @solnic's list.
The main problem with keeping things separate is toughness of management. I mean, it's not hard, but it will require some routine actions, such as maintaining CI builds, docs, separate issue trackers (this arguable, I admit), repo access, etc.
Don't get me wrong, I don't have a strong opinion on this topic, just don't want to see that ridiculous leftpad problem here :)

timriley:
Yeah, I think some discussion around the degree of packaging modularity we adopt would be good, since it'll be something we'd look to follow across the whole of the org.

Kind of related: isn't "dry-core" just "rom-support" by another name?

solnic:
> Kind of related: isn't "dry-core" just "rom-support" by another name?

Partly, yes. Some APIs will probably go back to `rom` (like `ArrayDataset` or `EnumerableDataset` and `DataProxy`), then there are things that should not be rom-specific, like `Inflector` with pluggable backends, or `Deprecations`. Then we have stuff that we probably no longer use, like `Publisher`, that could be removed. We also have some meta-programming helpers, like `ClassBuilder`, that is generic, fits better as a dry lib. Depending on how you look at it, we may also end up with everything from rom-support distributed across multiple dry gems, where one "dry-core" would ship a couple of APIs and probably a few additional, standalone gems, would be created too.

I'll write a summary of rom-support and post it here.

solnic:
rom-support summary:

* data_proxy - extension for enumerable objects that want to maintain their original type (ie `MyCollection#map` returns another `MyCollection` instance)
* array_dataset - a data proxy for objects wrapping an array
* enumerable_dataset - a data proxy for objects wrapping an enumerable
* auto_curry - method decorator that adds auto-currying (used by rom relations, could be moved back to rom core unless we think it can be made a generic API, hard to say tbf)
* cache - a helper for `Concurrent::Map` works with classes and objects and adds `cache.fetch_or_store` API
* class_builder - meta-programming API for generating classes with specific names
* class_macros - this is just class vars with inheritance, should be renamed to something that better describes what it is
* constants - a collection of useful constants like `EMPTY_HASH` etc.
* deprecations - APIs for deprecating methods with pluggable logging
* inflector - a generic inflection API with pluggable backends
* options - defines a constructor which accepts options as the last arg, supports default values, and type checking, this is identical to `dry-initializer` but smaller, I wonder if we could use `dry-initializer` instead assuming it's fast enough (this API is used EVERYWHERE in rom)
* registry - generic hash-based registry object, this should be moved back to rom core I think, and in rom 3.0 I'd like to see if we could use dry-container instead of this one
* inheritance_hook - I don't think we use it (????)
* publisher - I don't think we use it (????)

solnic:
> I'm not keen on the idea of dry-core, would rather have separate gems for each and keep each concern separate and independently usable, WDYT?

Yes and no. There's a serious cost that comes with each new gem that we add. Both from maintenance and usage povs. If we suddenly add 10+ new dry gems, the total cost will be significant, and if most of these gems are literally a single module or a single class, then we're gonna deal with a lot of additional friction that comes with maintenance, user support and so on.

It's probably sensible to say that if a given set of APIs is used almost-everywhere, then it should be merged into a single gem. Stuff like `equalizer` (a single module) or `options` (a single module) or `class_builder` (a single class) or cache (a single module) could be a single gem, its size would still be very small, and maintenance cost will be ~2-3x smaller than if these were separate gems.

I prefer small libs over micro libs, and I would draw a line between them by using a very simple approach - if a library is a single class or a single module, then it's a micro library (assuming it's a small class/module ie < 200LOC). This means that ie `dry-equalizer` is a micro lib, but `dry-auto_inject` is not (even though it started as one). It's great that dry-equalizer has been stable and we rarely do anything with it, but there's a flip side - since it's out of sight for most of the time, it doesn't get proper attention, if it was part of a gem collecting micro libs, chances are somebody would improve it from time to time.

Anyhow, just wanted to say that I would recommend starting with a single gem with all the micro stuff bundled in *and see how it evolves*. We can always extract things later into standalone gems.

timriley:
@solnic’s sold me on trying a combined approach for these things :)

flash-gordon:
@AMHOL what do you think? :)

AMHOL:
Bundling them and seeing how it evolves from there works for me :+1:

flash-gordon:
OK, I'm going to create a repo then and put some stuff there

solnic:
@flash-gordon can we close this then?

flash-gordon:
@solnic, sure, feel free

PR #133, opened 2016-08-27
Rename Hash#strict to Hash#permissive

backus:
As part of the planned `constructor_type` transition the `strict`
schema is being renamed to `permissive` to make room for a more
strict schema type.

See: https://gitter.im/dry-rb/chat?at=57686358feaf6cd222ad7e15

Also from gitter:

> @backus proposal looks good except :permissive which shouldn’t use defaults, so I just want it to work exactly the same as current :strict - this is really important for the common case of building structs from db data that you expect to be valid

### DIFF
--- a/lib/dry/types/hash.rb
+++ b/lib/dry/types/hash.rb
@@ -19,8 +19,8 @@ module Dry
         schema(type_map, Weak)
       end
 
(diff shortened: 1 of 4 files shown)


backus:
@solnic ready for review

PR #134, opened 2016-08-28
Add new Strict schema which rejects unknown keys

backus:


backus:
@solnic ready for review

lib/dry/types/hash/schema.rb
```diff
@@ -52,7 +52,26 @@ def call(hash, meth = :call)
...
+            end
+          end
+        end
+        alias_method :[], :call
+      end
+
+      class Strict < Schema
+        def call(hash, meth = :call)
+          unexpected = hash.keys - member_types.keys
+          raise UnknownKeyError.new(unexpected.first) unless unexpected.empty?
```
solnic: It would be better to provide all unexpected keys via `UnknownKeyError`, should ease debugging :) I know we fail on the first error in other places (ie constrained types in structs) but this is not how this should work, at the end of the day. I've had it in my TODO list for months and I'm planning to finally improve this in dry-struct.

backus: :thumbsup:

backus: @solnic fixed

PR #135, opened 2016-08-28
Improve coverage of Hash schemas

backus:
I'd like to refactor the code in #134 to reduce the massive duplication but the coverage doesn't seem to be great. I'm going to try to add tests and kill dead code in this PR first so that the subsequent refactor is less painful

### DIFF
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -81,7 +81,7 @@ module Dry
       class Weak < Schema
         def self.new(primitive, options)
           member_types = options.
-            fetch(:member_types, {}).
+            fetch(:member_types).
             each_with_object({}) { |(k, t), res| res[k] = t.safe }
 
           super(primitive, options.merge(member_types: member_types))
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -115,6 +115,8 @@ module Dry
         end
         alias_method :[], :call
       end
+
+      private_constant(*constants(false))
     end
   end
 end
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -88,7 +88,7 @@ module Dry
         end
 
         def try(hash, &block)
-          if hash.is_a?(::Hash)
+          if hash.instance_of?(::Hash)
             super
           else
             result = failure(hash, "#{hash} must be a hash")
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -4,7 +4,7 @@ module Dry
       class Schema < Hash
         attr_reader :member_types
 
-        def initialize(primitive, options)
+        def initialize(_primitive, options)
           @member_types = options.fetch(:member_types)
           super
         end
--- a/lib/dry/types/hash/schema.rb
+++ b/lib/dry/types/hash/schema.rb
@@ -12,7 +12,7 @@ module Dry
         def call(hash, meth = :call)
           member_types.each_with_object({}) do |(key, type), result|
             if hash.key?(key)
-              result[key] = type.__send__(meth, hash.fetch(key))
+              result[key] = type.public_send(meth, hash.fetch(key))
             else
               resolve_missing_value(result, key, type)
             end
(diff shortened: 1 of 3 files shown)


backus:
I'm not sure how to test the behavior provided by `Class` here: https://github.com/dry-rb/dry-types/blob/fix/hash_schema-coverage/lib/dry/types/hash.rb#L10

backus:
Alrighty the hash schemas are now 100% mutation covered.

lib/dry/types/hash.rb
```diff
@@ -12,13 +12,17 @@ def schema(type_map, klass = Schema)
             end
         }
 
-        klass.new(primitive, options.merge(member_types: member_types))
```
solnic: This way other options will be lost. This must not happen.

lib/dry/types/hash/schema.rb
```diff
@@ -4,15 +4,15 @@ class Hash < Definition
...
+        def initialize(_primitive, options)
           @member_types = options.fetch(:member_types)
           super
         end
 
         def call(hash, meth = :call)
           member_types.each_with_object({}) do |(key, type), result|
             if hash.key?(key)
-              result[key] = type.__send__(meth, hash[key])
+              result[key] = type.public_send(meth, hash.fetch(key))
```
flash-gordon: @backus I know you use it for semantics, but `.fetch` is slower than `.[]` https://github.com/JuanitoFatas/fast-ruby#hash-vs-hashfetch-code

lib/dry/types/hash.rb
backus: Will fix

lib/dry/types/hash/schema.rb
backus: @flash-gordon Noted. I do think fetch should still be used when possible though

backus:
@solnic options preservation resolved

lib/dry/types/hash/schema.rb
solnic: We don't need `fetch` here, though, given that we explicitly do `hash.key?(key)` one line above

backus: I like that it conveys to me that we are certain of the key being present. I agree it is weird to have under `key?` but this is an intermediary step (I did this PR so I could refactor some of the duplication in the schemas). Can we leave it for now?

solnic: One more thing - `public_send` is not needed here, parameterized `#call` is handled internally by the library, there's no need to be so careful, esp that it has a negative impact on the performance too.

backus: The semantics of `public_send` are more correct in this use case and the performance difference is not statistically significant

```
$ cat public-send.rb
require "benchmark/ips"

class Thing
  def method
  end
end

Benchmark.ips do |x|
  x.report("__send__")    { Thing.new.__send__(:method) }
  x.report("public_send") { Thing.new.public_send(:method)    }
  x.compare!
end

$ ruby public-send.rb
Warming up --------------------------------------
            __send__   259.922k i/100ms
         public_send   242.706k i/100ms
Calculating -------------------------------------
            __send__      5.311M (± 6.4%) i/s -     26.512M in   5.013065s
         public_send      4.854M (± 4.5%) i/s -     24.271M in   5.010566s

Comparison:
            __send__:  5311226.2 i/s
         public_send:  4854437.0 i/s - same-ish: difference falls within error
```

solnic: > Can we leave it for now?

Fair enough

> The semantics of public_send are more correct in this use case and the performance difference is just barely statistically significant

In an isolated benchmark, yes, but all those tiny little details add up easily. Let's stick to `public_send` for now and I'll look into total performance impact at the end, once we're done with refactorings, it may, or may not be an issue. We'll see :)

backus: > Let's stick to `public_send for now and I'll look into total performance impact at the end, once we're done with refactorings, it may, or may not be an issue. We'll see :)

Great thats all I really care about here. Basically: I'm cool with scrutinizing the refactored code for performance hotspots but this PR is just for me to get the code into a state that I'm confident in refactoring

solnic:
Looks like there are conflicts with master

backus:
@solnic Fixing

backus:
@solnic rebased

PR #137, opened 2016-08-28
Add definition specs for all types

flash-gordon:
I'm going to merge equality behavior specs into common definition specs further, then I'll move definition specs to lib dir so that they will be available in dry-struct /cc @solnic


backus:
@flash-gordon looks like you are improving similar things to what I was aiming to do in #135 so I've rebased my work on top of yours

flash-gordon:
OK, it's done I suppose. I tried to use shared specs in dry-struct, works fine except for complaining about lack of `.primitive`, but this is an easy thing to fix with `alias_method :primitive, :identity`

Issue #129, opened 2016-08-24
Surprising results for Form::Date.optional


padde:
Types::Form::Date works like this:

```
Types::Form::Date[nil]
# TypeError: no implicit conversion of nil into String
# from /usr/local/opt/rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/dry-types-0.8.1/lib/dry/types/coercions.rb:11:in `parse'

Types::Form::Date['']
#=> ""

Types::Form::Date['2016-01-01']
#=> Fri, 01 Jan 2016
```

but when we make the type optional in order to accept `nil` as value, it does no longer accept "invalid" date strings:

```
Types::Form::Date.optional[nil]
#=> nil

Types::Form::Date.optional['']
# Dry::Types::ConstraintError: "" violates constraints (#<struct Dry::Types::Result::Failure input="", error="\"\" must be an instance of Date">)
# from /usr/local/opt/rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/dry-types-0.8.1/lib/dry/types/sum.rb:40:in `block in call'

Types::Form::Date.optional['2016-01-01']
#=> Fri, 01 Jan 2016
```

Is this intentional or a bug? What would be the appropriate way to implement an optional Form::Date?

solnic:
Oh that's interesting, it is intentional since `optional` helper simply does `Types::Strict::Nil | Types::Form::Date` but we could make it smarter. For now this will work for you:

``` ruby
irb(main):008:0> (Types::Form::Nil | Types::Form::Date)['']
=> nil
irb(main):009:0> (Types::Form::Nil | Types::Form::Date)['2016-05-04']
=> #<Date: 2016-05-04 ((2457513j,0s,0n),+0s,2299161j)>
```

The reason why it works like that is that `Types::Form::Nil` knows that an empty string should be coerced to `nil`.

Anyhow, I'm gonna leave this open as we could improve `Type#optional` behavior.

solnic:
Actually I reported a new issue about this #130

padde:
@solnic thanks, but my use case is actually not with empty strings, but rather with invalid user input:

```
[89] pry(main)> (Types::Form::Nil | Types::Form::Date)['foo']
Dry::Types::ConstraintError: "foo" violates constraints (#<struct Dry::Types::Result::Failure input="foo", error="\"foo\" must be an instance of Date">)
[...]

PR #138, opened 2016-08-30
Graceful form types

padde:
Addresses the issues (but does not fix) in #129.

I added some specs for other form types as well, in order to ensure they keep failing gracefully in the future. One odd thing I experienced is that adding a similar test to the `form.nil | form.int` check will raise an error, so I left it out for now:

```
Failures:

  1) Dry::Types::Definition form.nil | form.int returns original value when it is not a string
     Failure/Error: raise ConstraintError, result

     Dry::Types::ConstraintError:
       #<Object:0x007f8b7d5414a0> violates constraints (#<Dry::Types::Result::Failure input=#<Object:0x007f8b7d5414a0> error="#<Object:0x007f8b7d5414a0> must be an instance of Integer">)
     # ./lib/dry/types/sum.rb:48:in `block in call'
     # ./lib/dry/types/sum.rb:61:in `try'
     # ./lib/dry/types/sum.rb:47:in `call'
     # ./spec/dry/types/types/form_spec.rb:32:in `block (3 levels) in <top (required)>'
```

I assume this has to do with the way sum types check the validity. Maybe the validity check needs to be adjusted for form types? If this is something you would like me to tackle I'd be happy to try it, but I would probably need a few pointers in the right direction first.

### DIFF
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -8,20 +8,23 @@ module Dry
       end
 
       def to_date(input)
+        return input unless input.respond_to?(:to_str)
         Date.parse(input)
-      rescue ArgumentError, TypeError
+      rescue ArgumentError
         input
       end
 
       def to_date_time(input)
+        return input unless input.respond_to?(:to_str)
         DateTime.parse(input)
-      rescue ArgumentError, TypeError
+      rescue ArgumentError
         input
       end
 
       def to_time(input)
+        return input unless input.respond_to?(:to_str)
         Time.parse(input)
-      rescue ArgumentError, TypeError
+      rescue ArgumentError
         input
       end
 
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -9,19 +9,19 @@ module Dry
 
       def to_date(input)
         Date.parse(input)
-      rescue ArgumentError
+      rescue ArgumentError, TypeError
         input
       end
 
       def to_date_time(input)
         DateTime.parse(input)
-      rescue ArgumentError
+      rescue ArgumentError, TypeError
         input
       end
 
       def to_time(input)
         Time.parse(input)
-      rescue ArgumentError
+      rescue ArgumentError, TypeError
         input
       end
 
--- a/spec/dry/types/types/form_spec.rb
+++ b/spec/dry/types/types/form_spec.rb
@@ -37,6 +37,11 @@ RSpec.describe Dry::Types::Definition do
     it 'returns original value when it was unparsable' do
       expect(type['not-a-date']).to eql('not-a-date')
     end
+
+    it 'returns original value when it is not a string' do
+      object = Object.new
+      expect(type[object]).to eql(object)
+    end
   end
 
   describe 'form.date_time' do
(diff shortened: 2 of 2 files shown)


lib/dry/types/coercions.rb
```diff
@@ -8,18 +8,21 @@ def to_nil(input)
...
         DateTime.parse(input)
       rescue ArgumentError
         input
       end
 
       def to_time(input)
+        return input unless input.respond_to?(:to_str)
         Time.parse(input)
       rescue ArgumentError
         input
```
solnic: Please just add `TypeError` to the rescue clauses (that's how it's done for other types like `Int` etc.). This should cover all cases where it may fail.

padde: I did that earlier, but unfortunately it doesn't work across all Ruby implementations.

Failed build for Rubinius:
https://travis-ci.org/dry-rb/dry-types/jobs/156124427

Failed Build for JRuby:
https://travis-ci.org/dry-rb/dry-types/jobs/156124428 

Switched to current method in this commit, with explanation:
https://github.com/dry-rb/dry-types/pull/138/commits/b47bbc167b5a1f7cfa702d94f6e7db7785aa89bf

padde:
@solnic maybe I should add some test cases for "pass through", e.g. passing a `Date` object to `form.date` etc.? I guess that would not be covered by the current implementation.

Or should this be handled on a more general level? Along the lines of: "when the input is already of the primitive type we are trying to cast to, we don't even need to perform casting".

padde:
@solnic re sum types with nil (in description), have you given this a thought? It seems like the sum type performs some check that the coerced result matches one of the primitive types, but that would not be very useful for summing form types, as they fall back to the original input if coercion fails. It's the main reason that I currently have to write

```
Date = (Types::Form::Nil | Types::Form::Date | Types::String)
```

instead of just

```
Date = (Types::Form::Nil | Types::Form::Date)
```

which seems counterintuitive.

padde:
Hey @solnic have you had a chance to look at my comments?

solnic:
I’ll be looking into that in a few hours


On 2 September 2016 at 15:49:07, Patrick Oscity (notifications@github.com)
wrote:

Hey @solnic <https://github.com/solnic> have you had a chance to look at my
comments?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dry-rb/dry-types/pull/138#issuecomment-244379858>, or mute
the thread

solnic:
Sorry for being slow with this. I was 200% busy with dry-validation/dry-logic work. Anyhow, I forgot to mention something important - we have a special type called `Safe` that should be used in cases where you are OK with coercion failures becoming safe, so that they return original input. This is used ie in dry-validation where we **try to coerce** and we're perfectly OK with coercion failing, but we do not want to have exceptions, we simply want the original input back.

Here's how it works in case of a sum type with coercible form types:

``` ruby
irb(main):001:0> (Types::Form::Nil | Types::Form::Int).safe['']
=> nil
irb(main):002:0> (Types::Form::Nil | Types::Form::Int).safe['123']
=> 123
irb(main):003:0> (Types::Form::Nil | Types::Form::Int).safe['gsgas']
=> "gsgas"
```

solnic:
Closed via https://github.com/dry-rb/dry-types/commit/3043ba4e9345fff2545d69b5c1f8035dddadd22d

solnic:
I rebased/squashed/pushed to master. Thanks for helping with this. We'll be releasing new dry-types next week along with dry-struct.

padde:
@solnic awesome, thanks for your time :)

PR #142, opened 2016-09-09
Fix #schema constructor raising different error

backus:
All other hash schemas raise `SchemaError` except for the `:schema` constructor type. This change resolves this discrepancy


backus:
@solnic ready for review

PR #143, opened 2016-09-09
Refactor hash schemas

backus:
A bit cleaner I think :smile:. As or more performant as the previous implementation

PR #144, opened 2016-09-10
Fix strict schemas accepting nil values when defaults are specified

backus:
I can't believe I almost forgot this 😆. This is basically the entire reason I originally got involved in the hash schemas and dry-struct


backus:
This PR depends on #143 which refactors the schemas (and made this change less painful). @solnic if you have reservations about #143 I can always update this PR to not depend on the refactoring

backus:
@solnic this is ready for review

PR #147, opened 2016-09-16
Maybe type is now an opt-in extension

flash-gordon:
@solnic pls, have a look. Especially on that `resolve_missing_value` trick


flash-gordon:
it seems private constants don't work well enough on rbx

solnic:
It would be good to separate specs for this extension, so we'd need to put its suite under `spec/extensions/maybe` and extend spec task to run the whole suite (along with its additional specs) so that we can run the suite with and without the extension loaded. WDYT?

lib/dry/types/hash/schema.rb
```diff
@@ -149,8 +149,6 @@ def resolve(hash)
           end
         end
       end
-
-      private_constant(*constants(false))
```
backus: You can replace the splat with the actual list of constant names instead of removing it entirely

flash-gordon: this was a quick and dirty hack to see if the specs will pass on rubinius. Regardless of this, what's the point of making constants private? Before this I didn't even know it's possible :) Shouldn't we just document them as private?

flash-gordon:
@solnic I agree completely, I made the PR to let you merge it and release a new gem version _before_ reworking specs structure, because this is a dry-v dependency and the later more or less is ready to go afaik. I'm going to work on specs during this day

lib/dry/types/hash/schema.rb
backus: Well why do we make methods private instead of just documenting them as private API? In both cases it is a signal to other collaborators and to end users that the method (or constant) is an implementation detail that shouldn't be interacted with directly.

flash-gordon: This is an absolutely valid point, but I'm asking because I suspect this method _is not_ so common as `Module#private`, and I just don't know how to use it properly. I checked dependencies in one of my projects and found out that constant hiding most often used in `concurrent-ruby` gem, but those are Java guys, so no surprises. Just kidding :) Anyway, there is no absolute privacy in ruby, you always can call a method, no matter if it is protected or private, but we _used to_ mark them as private in certain cases. I do not know anything about such cases for constants, so I probably won't use them to not introduce unexpected behaviors.
I'm going to have some fun with rubinius, then I'll be able to say if we can or cannot mark that `Schema` constant as private.

backus: Alrighty. I played with rubinius and found that you are able to do `private_constant(:Foo, :Bar)` but my trick to do `private_constant(*constants(false))` doesn't work (the latter privatizes all constants already defined in the current scope). 

Anyways yes nothing is truly private in ruby and even with private constants you can still do `const_get`. Ping me if you think there is no way to keep them private. I put work into these classes and wanted to clarify their interface too which is why I chimed in here.

solnic:
I never used this feature even though I knew it exists. I'm not against it
but if it causes issues then it doesn't feel like it's worth the effort tbh.

backus:
>I never used this feature even though I knew it exists. I'm not against it but if it causes issues then it doesn't feel like it's worth the effort tbh.


Shouldn't be an issue using the explicit version (`private_constant(:Foo)`) instead of the dynamic trick I was doing (`private_constant(*constants(false))`)

dry-types.gemspec
```diff
@@ -33,9 +33,10 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'dry-configurable', '~> 0.1'
   spec.add_runtime_dependency 'dry-logic', '~> 0.3', '>= 0.3.0'
   spec.add_runtime_dependency 'inflecto', '~> 0.0.0', '>= 0.0.2'
-  spec.add_runtime_dependency 'dry-monads', '>= 0.0.1'
+  spec.add_runtime_dependency 'dry-core', '>= 0.1.0'
 
   spec.add_development_dependency "bundler", "~> 1.6"
   spec.add_development_dependency "rake", "~> 11.0"
   spec.add_development_dependency "rspec", "~> 3.3"
+  spec.add_development_dependency 'dry-monads', '>= 0.1.1'
```
flash-gordon: not sure actually, there was an issue with 0.1. Anyway, I just released 0.2, so I'll bump the dep with `~> 0.2`

flash-gordon: also, I think I did write `>=` because this is a development dependency

lib/dry/types/hash/schema.rb
flash-gordon: @backus k, this wasn't hard https://github.com/dry-rb/dry-types/pull/147/files#diff-6f163e52d96701446ae4747de65b62ebR51 :)
I restored the initial version `private_constant(*constants(false)) `

flash-gordon:
@solnic I added a new rake task and configured rspec to run specs twice, take a look https://github.com/dry-rb/dry-types/pull/147/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR6

flash-gordon:
@backus fyi, that's what fails on rubinius and works on mri and jruby:
```ruby
module Foo
  module Bar
  end
  private_constant :Bar

  Bar # => Bar

  module Bar # => Private constant: Foo::Bar (NameError)
  end
end
```

backus:
Huh strange that rubinius has such different behavior

solnic:
Please just remove this. This PR is blocking the release and this constant visibility has very little value at the moment.

flash-gordon:
@solnic we already solved the issue, this was a side remark only :) feel free to merge it unless you have any further comments 
PS I'll file an issue in rubinius' tracker later today

solnic:
@flash-gordon thanks, could you update the changelog in master?

flash-gordon:
@solnic done 🎉 !

PR #150, opened 2016-09-23
Fixes typo in method spelling.

mitnal:
It should be Dry::Types.load_extensions(:maybe) instead of Dry::Types.load_extension(:maybe).
See https://github.com/dry-rb/dry-types/blob/master/Rakefile#L11


flash-gordon:
thanks!

PR #160, opened 2016-10-30
Allow to explicitly pass mapping to enum

bolshakov:
Allow to explicitly pass mapping between values and integer representation to enum. 

When working with legacy applications it may be useful to provide mapping between enum values and corresponding integer values.

``` ruby
Statuses = Types::Strict::String.enum('draft' => 0, 'published' => 10, 'archived' => 20)
Statuses[10] # => "published"
Statuses['published'] # => "published"
```

Passing values withot mapping works as well:

``` ruby
Statuses = Types::Strict::String.enum('draft', 'published', 'archived')
Statuses[1] # => "published"
Statuses['published'] # => "published"
```

Another problem is, lack of validation of integer values. So, you can create enum which maps Stings to Arrays for example.

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -33,7 +33,14 @@ module Dry
       end
 
       def enum(*values)
-        Enum.new(constrained(included_in: values), values: values)
+        mapping =
+          if values.length == 1 && values[0].is_a?(::Hash)
+            values[0]
+          else
+            ::Hash[values.zip(0...values.size)]
+          end
+
+        Enum.new(constrained(included_in: mapping.keys), mapping: mapping)
       end
 
       def safe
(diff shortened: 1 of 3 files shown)


bolshakov:
Folks, any concerns? Just tell me something

flash-gordon:
@bolshakov this is somehow missed, sorry about this. I'm OK with this but now it requires adding support at the AST level. Just right now I've found out we don't check if the resulting AST gets compiled back to the original type. I'll try to figure something out about it tomorrow.

@solnic do you know anything about why `visit_enum` is not there? It looks like we (mistakenly) skipped it.

flash-gordon:
ah, yeah, I added the missing part https://github.com/dry-rb/dry-types/commit/4cf170ed6d49379f0081b5015121f2eaa0b8f799 @bolshakov now this PR can be updated or just tell me and I'll do it myself

bolshakov:
@flash-gordon please do it yourself, I'm a bit out of scope now

flash-gordon:
FYI this wasn't merged into master yet, I just need to make some adjustments and think it through in general

PR #165, opened 2016-11-11
Added Y and N to TRUE_VALUES and FALSE_VALUES

scare21410:


fran-worley:
Do we have any tests for supported t/f values? If not should we add some?

solnic:
Yes [we do have tests](https://github.com/dry-rb/dry-types/blob/master/spec/dry/types/types/form_spec.rb#L86-L109) - this should be updated, maybe use constants to avoid duplication, it's kind of a smell to use details like that in a spec, but in this case it feels like a reasonable thing to do

solnic:
I merged this in and tweaked related spec to use the constants. Thanks for help.

PR #167, opened 2016-11-13
Reduce memory allocation with simple micro optimizations

davydovanton:
Hello,
I reduced some allocations. [As you can see][results] this changes improve benchmark about 25%.

[results]: https://gist.github.com/davydovanton/40ddc298496839edc3b11c2adc0086b1

### DIFF
--- a/lib/dry/types/definition.rb
+++ b/lib/dry/types/definition.rb
@@ -56,11 +56,11 @@ module Dry
       end
 
       def success(*args)
-        result(Result::Success, *args)
+        Result::Success.new(*args)
       end
 
       def failure(*args)
-        result(Result::Failure, *args)
+        Result::Failure.new(*args)
       end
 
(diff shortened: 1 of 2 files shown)


backus:
Could you also share the code that produced the benchmark?

davydovanton:
@backus it was scheme benchmark from dry-types repository :)

backus:
Ah ok. What about the allocation stuff?

solnic:
Aah nice. `each_with_object` is known to be slower

davydovanton:
@backus I used memory profiler (https://github.com/SamSaffron/memory_profiler) and call benchmark code in profiler block.

@solnic thanks for merge 🍻

solnic:
FYI this has been released already in v0.9.2. I noticed an improvement in perf in dry-validation too <3

PR #169, opened 2016-11-17
Fix form values specs

splattael:


flash-gordon:
😮 👍

solnic:
this is why we need mutation testing :(

PR #174, opened 2017-01-23
Returning a new instance of default value for default types

tchdp:
Ran into this bug when working on a project using dry-structs where the same instance of a default value is given to all instances of a class. It only applies when using defaults with a value `String.default('foo')`. 

The current workaround for this is to use the callable version `String.default { 'foo' }`. 
If this is expected behavior, I'd be happy to submit a PR to update the docs to reflect the nuance.

Use case where I ran into the issue:
```
require 'dry-struct'

describe 'Example' do
  module Types
    include ::Dry::Types.module
  end

  class User < ::Dry::Struct
    constructor_type :schema
    attribute :friends, ::Types::Array.default([])
  end

  it 'makes new friends' do
    user = User.new
    expect(user.friends.length).to eq 0 # passes

    user_2 = User.new
    user_2.friends << {}
    expect(user_2.friends.length).to eq 1 # passes

    # Fails here
    expect(user.friends.length).to eq 0
  end
end
```
The above example fails because `user` and `user_2` share the same reference to the default value `[]`
```
Failures:

  1) Example makes new friends
     Failure/Error: expect(user.friends.length).to eq 0

       expected: 0
            got: 1

       (compared using ==)
     # ./spec/example_spec.rb:24:in `block (2 levels) in <top (required)>'
```

It seems like `Dry::Struct` just delegates to the `Default` type for the value so I figured this was the correct spot to account for this change. I also updated the README since `rake spec` is not a defined task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `bundle exec rake spec` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
+After checking out the repo, run `bin/setup` to install dependencies. Then, run `bundle exec rake` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
 
 To install this gem onto your local machine, run `bundle exec rake install`. To release a new version, update the version number in `version.rb`, and then run `bundle exec rake release`, which will create a git tag for the version, push git commits and tags, and push the `.gem` file to [rubygems.org](https://rubygems.org).
 
--- a/lib/dry/types/default.rb
+++ b/lib/dry/types/default.rb
@@ -20,7 +20,11 @@ module Dry
       # @return [Object]
       attr_reader :value
 
-      alias_method :evaluate, :value
+      def evaluate
+        value.dup
+      rescue
+        value
+      end
 
       # @param [Object, #call] value
       # @return [Default, Dry::Types::Default::Callable]
--- a/spec/dry/types/default_spec.rb
+++ b/spec/dry/types/default_spec.rb
@@ -4,12 +4,21 @@ RSpec.describe Dry::Types::Definition, '#default' do
 
     it_behaves_like Dry::Types::Definition
 
-    it 'returns default value when nil is passed' do
-      expect(type[nil]).to eql('foo')
-    end
+    context 'when nil is passed' do
+      it 'returns default value' do
+        expect(type[nil]).to eql('foo')
+      end
 
-    it 'aliases #[] as #call' do
-      expect(type.call(nil)).to eql('foo')
+      it 'returns new instance of default value' do
+        ref_1 = type[nil]
+        ref_2 = type[nil]
+
+        expect(ref_1).not_to equal(ref_2)
+      end
+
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -20,7 +20,7 @@
 
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. Then, run `bundle exec rake spec` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
```
Kukunin: this change is wrong

Kukunin:
hm.. good catch! Just checked my project - found a couple of similar places.

`#dup` inside of default is kind of bad, because it's unexpected and implicit.
On other hand, it's so easy to get into this bug. 

If we consider Type as set of all possible values, every value is represented by its value itself (value object) and can be considered immutable.
With this said, we expect that `Dry::Types['strict.string']['foo'] == Dry::Types['strict.string']['foo']` 
Method `.[]` returns value, one from the whole set of possible values for such type. And this value is immutable.

So, one possible solution is to freeze argument given to `.default`. 
But `can't modify frozen String` exception isn't clear as well, and implicit as well - we need a better way to notify programmer about this nuance. 

But @tchdp's usecase is still valid, so we need to do something with it. Interesting to hear other ideas.

P.S. One more subtle bug caused by mutability

solnic:
If you want to mutate default value then it shouldn't be surprising that this side effect will affect all places where this value is used. It's a nasty gotcha but it only shows why mutating objects sucks. I don't it's a good idea to expect dry-types to dup defaults "just in case".

Kukunin:
@solnic let's think about good way to notify programmer about such nuance.

At least, describe such case in the documentation. 
It'd also may make sense to add `.frozen` scope to dry-type, which will be an alias to `.constructor(&:freeze)` - having such modifier would remind a programmer about mutating problems

flash-gordon:
@Kukunin it's not a good idea to have such an implicit behavior fwiw. This caveat definitely must be described in the docs. @solnic we could issue a warning if the value passed as a default is not frozen, I'm not sure about this decision but at least we could try and see whether it works fine or not

Kukunin:
@flash-gordon I didn't mean to freeze types by default, but just provide a neat ability to do so. 

    attribute :friends, ::Types::Array.default([]).frozen

Currently it's possible by 

    attribute :friends, ::Types::Array.default([]).constructor(&:freeze)

solnic:
Sorry but we're not going to do anything special about this. Thanks for PR and discussion. Closing this.

solnic:
ps. we definitely want to document this though :)

flash-gordon:
btw if anyone will program in Python, don't be surprised! :)
```ruby
$ python
Python 3.6.0 (default, Feb 14 2017, 01:02:14)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(a=[]):
...     a.append(1)
...     return a
...
>>> foo()
[1]
>>> foo()
[1, 1]
>>>
```

tchdp:
Thanks for taking the time to look through the issue, y'all. 👍

PR #175, opened 2017-01-24
Add Object type

### LINKED ISSUE #123: Add Types::Object type
This allows to accept any object, and constraint to objects of some specific class. Like:

``` ruby
MyClassType = Types::Object.constrained(type: MyClass)
```

flash-gordon:
fixes #123

It's a small useful addition allows you to have a placeholder type that can be later constrained with an actual type check. It's also not so uncommon in a dynamically typed language to be in a situation where you just don't know beforehand an object of what type you're going to have. For instance, SQLite despite being a SQL database can return you data of any type it supports in any field of any row, and we have to deal with it.

/cc @solnic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# to-be-released
+
+## Added
+
+* Added `Types::Object` which passes an object of any type (flash-gordon)
+
 # v0.9.3 2016-12-03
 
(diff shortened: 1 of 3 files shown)


flash-gordon:
refs https://github.com/rom-rb/rom-sql/issues/133

PR #178, opened 2017-02-16
Add type AST

solnic:
This adds `#to_ast` methods too all types. We need this in order to be able to dump types into ASTs and transform them into slightly different types. ie in rom-rb we'll be able to take database schema attribute types, and convert them to different types that are more suitable for other contexts, like struct attributes.

### TODO

- [x] `Definition#to_ast`
- [x] `Sum#to_ast`
- [x] `Constrained#to_ast`
- [x] `Constructor#to_ast`
- [x] `Hash#to_ast`
- [x] `Hash::Schema#to_ast`
- [x] `Hash::Permissive#to_ast`
- [x] `Hash::Strict#to_ast`
- [x] `Hash::StrictWithDefaults#to_ast`
- [x] `Hash::Weak#to_ast`
- [x] `Hash::Symbolized#to_ast`
- [x] `Array#to_ast`
- [x] `Array::Member#to_ast`
- [x] `Enum#to_ast`
- [x] `Safe#to_ast`
- [x] Tweak existing compiler to work with the AST (currently it relies on ASTs that include type identifiers from the builtin registry, we may want to make that available via `meta` so that when types are registered we set ie `meta(identifier: 'strict.int')` and then meta data can be included in ASTs so that compiler can reuse existing built-in types

### DIFF
--- a/lib/dry/types/array/member.rb
+++ b/lib/dry/types/array/member.rb
@@ -48,6 +48,13 @@ module Dry
             block ? yield(failure) : failure
           end
         end
+
+        # @api public
+        #
+        # @see Definition#to_ast
+        def to_ast
+          [:array, [:member, [member.to_ast]]]
+        end
       end
     end
   end
--- a/lib/dry/types/array/member.rb
+++ b/lib/dry/types/array/member.rb
@@ -53,7 +53,7 @@ module Dry
         #
         # @see Definition#to_ast
         def to_ast
-          [:array, [:member, [member.to_ast]]]
+          [:array, member.to_ast]
         end
       end
     end
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -16,19 +16,20 @@ module Dry
       end
 
       def visit_constructor(node)
-        primitive, fn = node
+        definition, fn = node
+        primitive = visit(definition)
         Types::Constructor.new(primitive, &fn)
       end
 
-      def visit_type(node)
-        type, args = node
-        meth = :"visit_#{type.tr('.', '_')}"
+      def visit_safe(node)
+        type, args = node[0]
+        Types::Safe.new(send(:"visit_#{type}", args))
+      end
 
-        if respond_to?(meth) && args
-          send(meth, args)
-        else
-          registry[type]
-        end
+      def visit_definition(node)
+        type, args = node
+
+        registry[args]
       end
 
       def visit_sum(node)
@@ -39,45 +40,19 @@ module Dry
         registry['array'].member(call(node))
       end
 
-      def visit_form_array(node)
-        registry['form.array'].member(call(node))
-      end
-
-      def visit_json_array(node)
-        registry['json.array'].member(call(node))
-      end
-
       def visit_hash(node)
         constructor, schema = node
         merge_with('hash', constructor, schema)
       end
 
-      def visit_form_hash(node)
-        if node
-          constructor, schema = node
-          merge_with('form.hash', constructor, schema)
-        else
-          registry['form.hash']
-        end
-      end
-
-      def visit_json_hash(node)
-        if node
-          constructor, schema = node
-          merge_with('json.hash', constructor, schema)
-        else
-          registry['json.hash']
-        end
-      end
-
-      def visit_key(node)
+      def visit_member(node)
         name, types = node
         { name => visit(types) }
       end
 
       def merge_with(hash_id, constructor, schema)
         registry[hash_id].__send__(
-          constructor, schema.map { |key| visit(key) }.reduce({}, :merge)
+          constructor, schema.first.map { |key| visit(key) }.reduce({}, :merge)
         )
       end
     end
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -16,8 +16,8 @@ module Dry
       end
 
       def visit_constructor(node)
-        definition, fn_object_id = node
-        fn = Dry::Types::FnContainer["fn_#{fn_object_id}"]
+        definition, fn_register_name = node
+        fn = Dry::Types::FnContainer[fn_register_name]
         primitive = visit(definition)
         Types::Constructor.new(primitive, &fn)
       end
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -16,7 +16,8 @@ module Dry
       end
 
       def visit_constructor(node)
-        definition, fn = node
+        definition, fn_object_id = node
+        fn = Dry::Types["fn_#{fn_object_id}"]
         primitive = visit(definition)
         Types::Constructor.new(primitive, &fn)
       end
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -28,7 +28,7 @@ module Dry
       end
 
       def visit_definition(node)
-        _, primitive = node
+        primitive = node
 
         if registry.registered?(primitive)
           registry[primitive]
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -27,9 +27,13 @@ module Dry
       end
 
       def visit_definition(node)
-        type, args = node
+        _, primitive = node
 
-        registry[args]
+        if registry.registered?(primitive)
+          registry[primitive]
+        else
+          Definition.new(primitive)
+        end
       end
 
       def visit_sum(node)
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -23,7 +23,7 @@ module Dry
       end
 
       def visit_safe(node)
-        type, args = node[0]
+        type, args = node
         Types::Safe.new(send(:"visit_#{type}", args))
       end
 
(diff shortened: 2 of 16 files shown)


lib/dry/types/hash/schema.rb
```diff
@@ -57,8 +57,18 @@ def try(hash, &block)
           end
         end
 
+        def to_ast
+          [:hash,
+           [hash_type,
+            [:member_types, [member_types.values.map(&:to_ast)]]]]
```
solnic: We also need names, so `member_types.map { |name, type| [:member, [name, type.to_ast]] }` should be OK :)

GustavoCaso: Ok great thanks

lib/dry/types/constructor.rb
```diff
@@ -71,6 +71,13 @@ def constrained_type
         Constrained::Coercible
       end
 
+      # @api public
+      #
+      # @see Definition#to_ast
+      def to_ast
```
GustavoCaso: I have some questions regarding the `Constructor` `#ast`, since it have `fn` that tells us information regarding the `coercion` method, I do not know if is important to included in the `ast`. Something like `[[:type, type.to_ast], [:fn, fn.to_s]]` id not really clear if we would be able to extract info from the `fn`

spec/dry/types/to_ast_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+      end
+    end
+  end
+
+  context 'Enum' do
+    subject(:type) { Dry::Types['strict.string'].enum('draft', 'published', 'archived') }
+
+    specify do
+      expect(type.to_ast).
+        to eql([
```
GustavoCaso: All the indentation is intended to be for readability, maybe after approving it, we can make it more compact, will commenting the PR I think is a good idea to see what `ast` are we getting

GustavoCaso:
@solnic I added the missing `ast` I left some comments, so feel free to add anything.

I think it will be more clear once we start working on the compiler 😄

lib/dry/types/array/member.rb
```diff
@@ -48,6 +48,13 @@ def try(input, &block)
             block ? yield(failure) : failure
           end
         end
+
+        # @api public
+        #
+        # @see Definition#to_ast
+        def to_ast
```
GustavoCaso: Here I was having doubts regarding if the better `ast` would be `[:hash, [:member, [:something, member.to_ast]]]` but I could come with a new apart from member

lib/dry/types/constructor.rb
solnic: for now just pass the fn as-is. @flash-gordon has an idea that we could register fn's explicitly in a registry, so that we can identify them easily. Having procs in AST is an anti-practice, but for now it's OK. We can improve it later once most of AST and a compiler are done.

lib/dry/types/array/member.rb
solnic: This will become clearer once we get to the compiler. We may realize that `:member` node is node even needed. Don't worry about this for now.

spec/spec_helper.rb
```diff
@@ -28,7 +28,13 @@ def call(_subject)
 Dir[Pathname(__dir__).join('shared/*.rb')].each(&method(:require))
 require_relative '../lib/spec/dry/types'
 
+Undefined = Dry::Core::Constants::Undefined
+
 RSpec.configure do |config|
+  config.before(:example, :maybe) do
```
GustavoCaso: I have no idea why the order matters, but `specs` where failing with `Dry::Container::Error:
Nothing registered with the key "maybe.strict.string"`

spec/dry/types/compiler_spec.rb
```diff
@@ -217,18 +170,18 @@
   end
 
   it 'builds an schema-less form.hash' do
-    ast = [:type, ['form.hash']]
+    ast = Dry::Types['form.hash'].schema([]).to_ast
 
     type = compiler.(ast)
 
-    expect(type[nil]).to be(nil)
+    expect(type[nil]).to eql({})
```
GustavoCaso: This was the only spec I could not fulfill as it was, the coercion was returning always a `hash`, to code that makes this condition to happen is here https://github.com/dry-rb/dry-types/blob/type-ast/lib/dry/types/hash/schema.rb#L101-L102 since is an empty schema it did not have any `members` I don't have that concrete knowledge of the desired behavior

lib/dry/types/hash/schema.rb
```diff
@@ -57,8 +57,24 @@ def try(hash, &block)
...
         end
 
+        def to_ast
+          [
+            :hash,
+            [
+              hash_type,
+              [
+                member_types.map { |name, member| [:member, [name, member.to_ast]] }
+              ]
```
GustavoCaso: I decided to remove the [:member_types, ...] to make compiler easier to work

lib/dry/types/compiler.rb
```diff
@@ -39,45 +40,19 @@ def visit_array(node)
...
-      def visit_key(node)
+      def visit_member(node)
         name, types = node
         { name => visit(types) }
       end
 
       def merge_with(hash_id, constructor, schema)
         registry[hash_id].__send__(
-          constructor, schema.map { |key| visit(key) }.reduce({}, :merge)
+          constructor, schema.first.map { |key| visit(key) }.reduce({}, :merge)
```
GustavoCaso: `first` is the only thing that annoys me

GustavoCaso:
At the end I decided to remove the `to_ast` from `Array` and `Hash`, since the inherit the one from `Definition`

lib/dry/types/compiler.rb
```diff
@@ -11,24 +13,30 @@ def call(ast)
...
-          registry[type]
-        end
+      def visit_definition(*node)
+        primitive, meta = node
+        definition = if registry.registered?(primitive)
+                       registry[primitive]
+                     else
+                       Definition.new(primitive)
+                     end
+        meta.any? ? definition.meta(meta) : definition
```
solnic: This could be just `meta ? definition.meta(meta) : definition`

GustavoCaso: I think even if you don't have any `meta` by default it would by an empty hash, since we added meta as part of the `ast` https://github.com/dry-rb/dry-types/pull/178/files#diff-58177960b3ed930fac3ecb590a2d3f35R108

lib/dry/types/compiler.rb
solnic: Yes it's a smell, let's get rid of this :) I *think* we could have `[:hash, [:strict, [[:member, [:age, [:definition, Integer]]]]]]` then you can just do `constructor, members = node` and visit each member and create a schema from the result.

lib/dry/types/compiler.rb
solnic: Ah good point, so please do `meta.empty?` as it's faster :)

GustavoCaso: Great 👍

lib/dry/types/compiler.rb
GustavoCaso: You mention about getting rid of `[:hash, [:strict, [..]]]` in favor of `[:hash_strict, [...]]`

solnic: Right, but the hash visitor makes sense as it is now, so let's leave it like that

flash-gordon:
@solnic anything else? :)

lib/dry/types/enum.rb
```diff
@@ -36,6 +36,13 @@ def call(input)
         type[value || input]
       end
       alias_method :[], :call
+
+      # @api public
+      #
+      # @see Definition#to_ast
+      def to_ast
+        [:enum, [type.to_ast]]
+      end
```
solnic: This should be `[:enu, type.to_ast]` extra nesting is redundant

lib/dry/types/safe.rb
```diff
@@ -33,6 +33,13 @@ def try(input, &block)
         block ? yield(result) : result
       end
 
+      # @api public
+      #
+      # @see Definition#to_ast
+      def to_ast
+        [:safe, [type.to_ast]]
+      end
```
solnic: Same problem here, this extra nesting shouldn't be needed

lib/dry/types/hash/schema.rb
solnic: ...and here too, this could be `[:hash, [hash_type, member_types.map { ... }]]`

solnic:
@GustavoCaso @flash-gordon I left a couple comments, all related to the same small issue

GustavoCaso:
@flash-gordon @solnic I apply the changes hope everything is as expected 😄

flash-gordon:
I rebased and force-pushed the branch

solnic:
I think this is good-enough to be merged, there's just one caveat - it breaks dry-validation. I propose creating `release-0.10` branches in dry-v and dry-t and start treating master as a bleeding edge targeting 0.11 releases. WDYT?

flash-gordon:
yeah, sounds like a plan, I'll use this branch to start the work on dumping `rom` schemas to ast

solnic:
@flash-gordon OK, let's get this party started then :)

solnic:
This is now merged. Thank you @GustavoCaso for this effort and let's continue polishing/finishing stuff for 0.11. I'll cut a release-0.10 branch too for the ongoing maintenance.

PR #182, opened 2017-03-30
Make YARD types correct

backus:
I have revised these YARD annotations using a tool that instruments your
tests to check types: [yardcheck](https://github.com/backus/yardcheck).

These are the changes I needed to make to make the tool pass.

This is still a WIP because I am trying to also use the great YARD docs in dry-types to help make the tool's recommendations "natural" if that makes sense.

### DIFF
--- a/lib/dry/types/default.rb
+++ b/lib/dry/types/default.rb
@@ -3,6 +3,7 @@ require 'dry/types/decorator'
 module Dry
   module Types
     class Default
+      include Type
       include Dry::Equalizer(:type, :options, :value)
       include Decorator
       include Builder
--- a/lib/dry/types/default.rb
+++ b/lib/dry/types/default.rb
@@ -24,7 +24,7 @@ module Dry
       alias_method :evaluate, :value
 
       # @param [Object, #call] value
-      # @return [Default, Dry::Types::Default::Callable]
+      # @return [Class] {Default} or {Default::Callable}
       def self.[](value)
         if value.respond_to?(:call)
           Callable
@@ -33,7 +33,7 @@ module Dry
         end
       end
 
-      # @param [Definition] type
+      # @param [Type] type
       # @param [Object] value
       def initialize(type, value, *)
         super
@@ -52,7 +52,7 @@ module Dry
       end
 
       # @param [Object] input
-      # @return [Success]
+      # @return [Result::Success]
       def try(input)
         success(call(input))
       end
--- a/lib/dry/types.rb
+++ b/lib/dry/types.rb
@@ -12,6 +12,7 @@ require 'dry/core/constants'
 
 require 'dry/types/version'
 require 'dry/types/container'
+require 'dry/types/type'
 require 'dry/types/definition'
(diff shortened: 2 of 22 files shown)


backus:
@alsemyonov I'd love if you could review this since you added most of the great YARD documentation.

backus:
Feedback from @flash-gordon which I'll update this with later:

> Nikita Shilnikov @flash-gordon 11:40
> @backus that's nice. However I tend to whink Type is rather a module than a class because a type is what acts as a type, not what is a type
> 
> John Backus @backus 11:43
> Not sure how that would work in a way that supports extracting an abstract interface for every kind of type
> 
> Nikita Shilnikov @flash-gordon 11:46
> @backus I mean we because of how inheritance implemented in Ruby requiring Type to be a class is too strong FWIW, when it's a module you always can mixin it
> examples are dry-struct and Attribute class in rom
> 
> John Backus @backus 11:52
> Ah ok I get what you mean
> The tool supports that so I can make that change

lib/dry/types/default.rb
```diff
@@ -23,7 +24,7 @@ def evaluate
       alias_method :evaluate, :value
 
       # @param [Object, #call] value
-      # @return [Default, Dry::Types::Default::Callable]
+      # @return [Class] `Default` or `Callable`
```
alsemyonov: When I wrote this, I have tried to make documentation more helpful and provide links to returned classes.
Unfortunately, AFAIK [YARD](http://yardoc.org/) does not support putting specific `Class` instances returned from a method in `@return` tag (because the class name in `@return` block means “instance of the class”). 

Following that direction, to have links to those classes, I think, It is probably better to link to the actual classes, and not only mark them as code in `@return` so following would be more helpful:

```ruby
# @return [Class] {Default} or {Default::Callable}
```

backus: Great idea!

backus: > Unfortunately, AFAIK YARD does not support putting specific Class instances returned from a method in @return tag (because the class name in @return block means “instance of the class”).

Yeah. I'd like some form of syntax for this eventually. I opened https://github.com/backus/yardcheck/issues/6 so I can start thinking about solutions for Yardcheck

backus:
@alsemyonov really glad you like the tool! I fixed the regression you pointed out :smile:

lib/dry/types/default.rb
alsemyonov: I was thinking about such feature in `YARD` some time ago. I would rather see that as a core `YARD` feature or a `YARD` plugin, extending `@return` (and maybe some other tags) syntax, so we have this functionality accessible both programmatically and in human-readable form. 

Then on top of it, it would be good to see corresponding `yardcheck` functionality.

If we could discuss and figure out the syntax that would fit this feature properly then later I could try to build such plugin and expose data needed for `yardcheck`.

backus:
Yardcheck supports errors now!

image

Pushed a few more fixes :smile:

PR #185, opened 2017-04-18
Do not store meta in options

flash-gordon:
@solnic I did this because atm `type != type.meta({})` and this is obviously a bug. We need to store an empty hash for meta in options or not to store meta in options at all. I think the latter is better, what do you think about it? 
This allows to simplify code [like this](https://github.com/hanami/model/blob/ade8bf454f08f0bb066dc878aa962f20f11af299/lib/hanami/model/sql/types.rb#L37-L56) at least, while we don't have AST-transformations for types.

### DIFF
--- a/lib/dry/types/hash.rb
+++ b/lib/dry/types/hash.rb
@@ -17,7 +17,7 @@ module Dry
             end
         }
 
-        klass.new(primitive, options.merge(member_types: member_types))
+        klass.new(primitive, options.merge(member_types: member_types, meta: meta))
       end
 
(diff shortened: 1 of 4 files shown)


solnic:
> or not to store meta in options at all

This is a good idea, let's do this

GustavoCaso:
Would this affect to the `AST` construction ?

flash-gordon:
@GustavoCaso nope, pretty much the opposite, we could replace simple equality checks with `AST` trimming or something like that, depending on the task

PR #186, opened 2017-04-22
Return original value when Date.parse raises a RangeError.

jviney:


solnic:
Nice catch. Thanks for fixing this.

PR #191, opened 2017-04-27
Add Type#optinal?

GustavoCaso:
This method returns a boolean indicating if type is `Sum` and `left` is `nil`

### DIFF
--- a/lib/dry/types/options.rb
+++ b/lib/dry/types/options.rb
@@ -26,11 +26,6 @@ module Dry
       def meta(data = nil)
         data ? with(meta: @meta.merge(data)) : @meta
       end
-
-      # @return [Boolean]
-      def optional?
-        self.class == Sum && left.primitive == NilClass
-      end
     end
   end
 end
--- a/lib/dry/types/options.rb
+++ b/lib/dry/types/options.rb
@@ -26,6 +26,11 @@ module Dry
       def meta(data = nil)
         data ? with(meta: @meta.merge(data)) : @meta
       end
+
+      # @return [Boolean]
+      def optional?
+        self.class == Sum && left.primitive == NilClass
+      end
     end
   end
 end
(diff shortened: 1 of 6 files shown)


lib/dry/types/options.rb
```diff
@@ -26,6 +26,11 @@ def with(new_options)
       def meta(data = nil)
         data ? with(meta: @meta.merge(data)) : @meta
       end
+
+      # @return [Boolean]
+      def optional?
```
flash-gordon: this should be split in two pieces, one is in `definition.rb` (look at [these methods](https://github.com/dry-rb/dry-types/blob/278b03588b1f1a9ccecf422f9c886c80a7f61bc5/lib/dry/types/definition.rb#L44-L52))
another is in `sum.rb`, this way we can avoid the `self.class == Sum` check

flash-gordon:
@GustavoCaso thanks man, we're almost there :)

GustavoCaso:
After making the pull request I thought about that. Thanks @flash-gordon

lib/spec/dry/types.rb
```diff
@@ -32,6 +32,12 @@ def be_boolean
       expect(type).to eq(type)
     end
   end
+
+  describe '#optional?' do
+    it 'returns a boolean value' do
+      expect(type.optional?).to be_boolean
+    end
```
solnic: More idiomatic way in rspec is to write `expect(type).to be_optional`

spec/dry/types/sum_spec.rb
```diff
@@ -11,6 +11,20 @@
...
+    it 'return true if left side is nil' do
+      type = Dry::Types['strict.nil'] | Dry::Types['string']
+
+      expect(type.optional?).to eq true
+    end
+
+    it 'return false if left is not nil' do
+      type = Dry::Types['string'] | Dry::Types['nil']
+
+      expect(type.optional?).to eq false
```
solnic: Similarly here: `expect(type).to_not be_optional`

PR #193, opened 2017-05-02
make .enum.default raise an error.

### LINKED ISSUE #189: Weird behaviour with `options` when defining an enum with a default
I've spotted some weird behaviour when defining an enum with a default:

```ruby
require 'dry-types'

module Types
  include Dry::Types.module
end

string = Types::Strict::String

default_first = string.default('foo').enum('foo', 'bar')
default_first.options
# => {:values=>["foo", "bar"]}
default_first.values
# => ["foo", "bar"]

default_last = string.enum('foo', 'bar').default('foo')
default_last.options
# => {}
default_last.values
# => NoMethodError Exception: undefined method `values' for #<Dry::Types::Default:0x007fdcaa82a170>
```

As you can see, calling `default` then `enum` works as expected, but calling `enum` then `default` means that the `options` hash won't be set correctly and `values` won't work.

It's not a big deal, because I can just use the 'default first' approach, but it's still unintuitive - I had no idea that the order mattered, and bugs resulted.

Not sure what the best way to fix this; I don't know enough about the internals of dry-types. Or maybe the problem's not worth fixing, in which case I think the docs should be updated to point out this pitfall. (Or `Enum#default` should raise an error and tell you to use `Default#enum` instead, rather than slyly returning a type that won't quite behave as expected.)

georgemillo:
Closes #189.

Wasn't sure whether to raise a custom error type or not - this PR just raises `RuntimeError`. Feel free to tweak the error message.

Note that this is a breaking change.

### DIFF
--- a/spec/dry/types/enum_spec.rb
+++ b/spec/dry/types/enum_spec.rb
@@ -29,10 +29,10 @@ RSpec.describe Dry::Types::Enum do
       expect(with_default[nil]).to eql('draft')
     end
 
-    it 'allows defining a default value for an enum' do
-      with_default = type.default('published')
-
-      expect(with_default[nil]).to eql('published')
+    it "doesn't allows defining a default value for an enum" do
+      expect do
+        type.default('published')
+      end.to raise_error(RuntimeError)
     end
 
     it 'aliases #[] as #call' do
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -36,6 +36,11 @@ module Dry
         type[value || input]
       end
       alias_method :[], :call
+
+      def default(*)
+        raise '.enum(*values).default(value) is not supported. Call '\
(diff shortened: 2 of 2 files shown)


spec/dry/types/enum_spec.rb
```diff
@@ -29,10 +29,10 @@
       expect(with_default[nil]).to eql('draft')
     end
 
-    it 'allows defining a default value for an enum' do
-      with_default = type.default('published')
-
-      expect(with_default[nil]).to eql('published')
+    it "doesn't allows defining a default value for an enum" do
```
Neodelf: I'm not sure if it's important. But сan not leave the agreement and use single quotes?

flash-gordon: tbf we aren't pedantic about this

flash-gordon:
thanks @georgemillo

PR #195, opened 2017-05-04
Added #of alias for Dry::Types::Array#member

maliqq:


solnic:
No strong feelings about this. WDYT @timriley?

flash-gordon:
I think this is fine :)

timriley:
I missed the notification back on 12th of May, but yeah, this seems innocuous.

My only question would be whether `#of` should be the official API, and `#member` deprecated? I don't really see much point in keeping _both_ methods around.

flash-gordon:
Yeah, from now on `#of` is my personal preference, @solnic wdyt?

solnic:
It's OK

PR #196, opened 2017-05-16
Pass result to schema error for improving error message

GustavoCaso:
These fix https://github.com/dry-rb/dry-struct/issues/16

By passing the `result` message to error we will be able to access it in `dry-struct`


solnic:
Thanks Gustavo <3

PR #200, opened 2017-06-03
Add meta to the AST

flash-gordon:
/cc @solnic 🎉


solnic:
Sorry but we gotta tweak it :) We want to have structures like this: `[:definition, [String, meta]]` so that each visitor receives a node, instead of two args. This way each visitor can decide what to do with meta.

flash-gordon:
You're right, I actually misinterpreted things when I was looking at how this was done in rom-repo, thanks for pointing out, will fix it right away

PR #202, opened 2017-06-12
Support for excluding meta from the AST representation

flash-gordon:


flash-gordon:
@solnic we don't want to dump all the stuff we have in meta to the tree in ROM, `:sql_expr` as an example

PR #203, opened 2017-06-22
Update compiler to work with dry-validations

solnic:


lib/dry/types/compiler.rb
```diff
@@ -53,14 +53,30 @@ def visit_hash(node)
         merge_with('hash', constructor, schema).meta(meta)
       end
 
+      def visit_json_array(node)
+        registry['json.array'].member(call(node))
+      end
+
+      def visit_json_hash(node)
+        if node
```
solnic: this doesn't look good, why can this be nil?

GustavoCaso:
Close in favor https://github.com/dry-rb/dry-types/pull/204

PR #205, opened 2017-06-23
Add :visit_constrained to compiler

GustavoCaso:
@solnic I add `visit_constrained` also `visit_and` and `visit_or` and we missing any more?


lib/dry/types/compiler.rb
```diff
@@ -16,6 +16,11 @@ def visit(node)
         send(:"visit_#{ type }", body)
       end
 
+      def visit_constrained(node)
+        definition, rule, meta = node
+        Types::Constrained.new(visit(definition), rule: visit(rule)).meta(meta)
```
solnic: It's better to do `Types.rule_compiler.(rule)` and remove `visit_or` and `visit_and`

GustavoCaso: Perfect. I will change it. I wasn't sure if that was going to work, because of the reduce(&:and) part. I thought each operation has to use a different reduce function

GustavoCaso:
@solnic are we good for merging this one?

PR #206, opened 2017-06-23
Update README to contain correct command to run specs

janpieper:
The README contains an outdated command to run the specs.

__Wrong__
```
bundle exec rake spec
```

__Correct__
```
bundle exec rake run_specs
```

There are at least these other commands, you can use to run the specs:

* `bundle exec rake` (because `run_specs` is set to be the default task)
* `bundle exec rspec`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


flash-gordon:
ah, right, thanks 👍

Issue #197, opened 2017-05-17
Add single-value types


flash-gordon:
In some cases, you want to have a type that has a single instance. For example, Ruby's `TrueClass`, `FalseClass`,  and `NilClass` all have a single value `true`, `false`, and `nil` correspondingly. To make it more generic we can have something like this

```ruby
# uses .equal? in valid?
nil_type = Types::Constant(nil)
true_type = Types::Constant(true)

# uses .== in valid?
empty_list_type = Types::Value([])
```

This can be implemented with `Constrained` but maybe we need more predicates in dry-logic for that.

@solnic WDYT?

solnic:
what'd be the difference between this and `Types::Strict::Nil` etc.?

flash-gordon:
I don't mean any difference for existing types, but it would allow building app-level constant types, e.g.

```ruby
Success = Types::Constant('success')
Failure = Types::Constant('failure')

Result = Success | Failure

case result
when Success then # happy path
when Failure then # 
end
```

PR #210, opened 2017-08-26
[WIP] Add type builders to auto-generated module

flash-gordon:
Related to https://github.com/dry-rb/dry-monads/pull/43, closes https://github.com/dry-rb/dry-types/issues/197

This adds convenient methods for building types using the auto-generated module. I implemented most common cases from my POV but I eager to hear any suggestions.

- `Array(type)` is equivalent to `Array.of(type)`
- `Hash(:symbolized, mapping)` is equivalent to `Hash.symbolized(mapping)`
- `Instance(class)` is equivalent to `Dry::Types::Definition.new(class).constrained(type: class)`, this one saves quite a bit of typing :)
- `Value(pattern)` returns a type that checks if a value is equal (using `eql?`) to the pattern
- `Constant(object)` returns a type that checks if a value is the object itself

@solnic thoughts?


solnic:
A shortcut for definition with a constructor would be nice too. We could use it in rom already and remove it from there.

flash-gordon:
Ahh, right, good point

Issue #184, opened 2017-04-12
dry/types/decorator.rb:74:in `method_missing': undefined method `failure' for #<Dry::Types::Safe> (NoMethodError)


joevandyk:
```
require 'dry-struct'

module Types
  include Dry::Types.module
end

class Contact < Dry::Struct
  constructor_type :symbolized
  attribute :age, Types::Coercible::Int.optional
end

Contact.new("age" => "2.2")
```

results in

```
/Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/decorator.rb:74:in `method_missing': undefined method `failure' for #<Dry::Types::Safe:0x007f969b85ef10> (NoMethodError)
Did you mean?  fail
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/safe.rb:31:in `rescue in try'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/safe.rb:29:in `try'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/safe.rb:13:in `call'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:75:in `block in coerce'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:217:in `block in resolve'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:208:in `each'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:208:in `resolve'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:73:in `coerce'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/hash/schema.rb:27:in `call'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-types-0.9.4/lib/dry/types/constructor.rb:39:in `call'
	from /Users/joe/.gem/ruby/2.3.3/gems/dry-struct-0.2.1/lib/dry/struct/class_interface.rb:77:in `new'
[...]

PR #212, opened 2017-09-29
Add success and failure method for Sum type

GustavoCaso:
#184 

After checking the code looks like the `Sum` type was missing the `success` method, creating some errors using `dry-struct`.

This PR adds the `success` method.

### DIFF
--- a/spec/dry/types/sum_spec.rb
+++ b/spec/dry/types/sum_spec.rb
@@ -120,6 +120,20 @@ RSpec.describe Dry::Types::Sum do
     end
   end
 
+  describe '#success' do
+    subject(:type) { Dry::Types['strict.bool'] }
+
+    it 'returns success when value passed' do
+      expect(type.success(true)).to be_success
+    end
+
+    it 'raises ArgumentError when non of the types have a vallid input' do
+      expect{
+        type.success('true')
+      }.to raise_error(ArgumentError, /Invalid success value for/)
+    end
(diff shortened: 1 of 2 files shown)


solnic:
Could you also add `failure` so that it can close the related issue too?

GustavoCaso:
@solnic Added!

GustavoCaso:
@solnic any update on this one?

solnic:
Almost there. We need a spec that ensures that the original issue is fixed too. Just make sure something like this works:

``` ruby
Types::Coercible::Int.optional.safe[nil] # should return a success
Types::Coercible::Int.optional.safe['1'] # should return a success
Types::Coercible::Int.optional.safe['foo'] # should return a failure
```

spec/dry/types/sum_spec.rb
```diff
@@ -120,6 +120,28 @@
...
+      expect{
+        type.success('true')
+      }.to raise_error(ArgumentError, /Invalid success value 'true'/)
+    end
+  end
+
+  describe '#failure' do
+    subject(:type) { Dry::Types['int'] | Dry::Types['string'] }
+
+    it 'returns success when value passed' do
```
solnic: This should be `it 'returns failure when invalid value is passed'`

GustavoCaso:
@solnic I added the correct spec description 😄

Issue #215, opened 2017-10-16
Types::Json::Decimal raises NoMethodError on nil


nesaulov:
Hello. First of all, thank you for your work. I think I have encountered unexpected behaviour: when `Types::Json::Decimal` is given explicit `nil` value, `NoMethodError` is raised, as opposed to `#<Dry::Types::Result::Failure input=nil error="nil must be an instance of BigDecimal">` which seems to be more natural (like when empty string is passed, for example).
Here is how to reproduce:
```ruby
Types::Json::Decimal.try(nil) 
NoMethodError: undefined method `to_d' for nil:NilClass
Did you mean?  to_c
               to_r
               to_h
               to_a
               to_i
               to_s
               to_f
        from /Users/nesaulov/Code/ruby/opensource/dry-types/lib/dry/types/coercions/json.rb:16:in `to_decimal'
        from /Users/nesaulov/Code/ruby/opensource/dry-types/lib/dry/types/constructor.rb:56:in `[]'
        from /Users/nesaulov/Code/ruby/opensource/dry-types/lib/dry/types/constructor.rb:56:in `try'
```
I have created a PR addressing this issue.

flash-gordon:
was fixed via #216

PR #216, opened 2017-10-16
Return `Dry::Types::Result::Failure` on `Types::Json::Decimal.try(nil)`

nesaulov:
#215 
@solnic @flash-gordon please review


flash-gordon:
thank you!

PR #228, opened 2017-11-21
Remove dry-configurable as dependency

GustavoCaso:
As we talk about. 

This removes `dry-configurable` as a dependency.


timriley:
Was the reason for this so that dry-configurable can add dry-types as a dependency of its own?

solnic:
@timriley that too, but mostly it was an overkill to rely on configurable just for a single class-level setting.

GustavoCaso:
@timriley sorry for the lack of information in the PR, next time I will write more context 😄

PR #230, opened 2017-12-16
Rework hash schemas

flash-gordon:
This branch depends on https://github.com/dry-rb/dry-types/pull/211

This is a promising start, the end goal is discarding hash schemas. I removed most of the code that relies on inheritance, the next step is removing inheritance altogether. I plan to provide backward compatibility for hash schemas, however, the main point is to provide enough flexibility for building custom-behavior schemas _without_ introducing a ton of barely compatible options. At the moment, I think we'll have to separate weak/symbolized schemas from other types.

### DIFF
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -25,15 +25,19 @@ module Dry
 
       # @param [Object] input
       # @return [Object]
-      def call(input)
+      def call(input = Undefined)
         value =
-          if values.include?(input)
+          if input == Undefined
+            type.call
+          elsif values.include?(input)
             input
           elsif mapping.key?(input)
             mapping[input]
+          else
+            input
           end
 
-        type[value || input]
+        type[value]
       end
       alias_method :[], :call
 
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -27,7 +27,7 @@ module Dry
       # @return [Object]
       def call(input = Undefined)
         value =
-          if input == Undefined
+          if input.equal?(Undefined)
             type.call
           elsif values.include?(input)
             input
--- a/lib/dry/types/hash/schema_builder.rb
+++ b/lib/dry/types/hash/schema_builder.rb
@@ -1,3 +1,5 @@
+require 'dry/types/hash/schema'
+
 module Dry
   module Types
     class Hash < Definition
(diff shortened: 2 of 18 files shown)


flash-gordon:
@solnic this is ready to go to master, I replaced all the schemas we have with a schema builder for backward compatibility (and besides that, it was easier than rewriting from scratch). The PR doesn't break anything beyond https://github.com/dry-rb/dry-types/pull/211 which is a reasonable breaking change for default types. My plan is to provide a new flexible yet concise and simple API for building schemas. I'd also deprecate the old schemas as well

lib/dry/types/hash/schema_builder.rb
```diff
@@ -4,15 +4,16 @@
...
       class SchemaBuilder
-        def initialize
-          @nil_to_undefined = -> v { v.nil? ? Undefined : v }
-          @omittable_keys = %i(schema weak symbolized).freeze
-          @permissive = %i(schema permissive weak symbolized).freeze
-
-          freeze
-        end
+        NIL_TO_UNDEFINED = -> v { v.nil? ? Undefined : v }
+        OMITTABLE_KEYS = %i(schema weak symbolized).freeze
```
solnic: could we rename this to `OPTIONAL_KEYS`? it'd be more consistent with dry-v's `required/optional`

lib/dry/types/enum.rb
```diff
@@ -25,15 +25,19 @@ def initialize(type, options)
 
       # @param [Object] input
       # @return [Object]
-      def call(input)
+      def call(input = Undefined)
         value =
-          if values.include?(input)
+          if input == Undefined
```
solnic: I think we want `input.equal?(Undefined)` here, just like in `Default#call`

lib/dry/types/hash/schema_builder.rb
flash-gordon: @solnic I don't have any preference about naming but this comes from `meta[:omittable]` where it means whether a key is optional. The problem is that this information is stored inside the type itself meaning if we change it from `:omittable` to `:optional` then we end up with something like

```ruby
age: Dry::Types['strict.int'].optional.meta(optional: true)
```
when we want have a key that can be omitted _and_ can have `nil` value. This looks rather confusing, don't you think? We don't have a separate place for key meta and I'd not go this way, yet.

PR #231, opened 2018-01-11
New API for hash schemas

flash-gordon:
This adds the following API for hash schemas:

1. `Schema#with_key_transform`—transforms keys of input hashes, for things like symbolizing etc.
1. `Schema#strict`—makes a schema intolerant to unknown keys.
1. `Hash#with_type_transform`—transforms member types with an arbitrary block. For instance, 
```ruby
optional_keys = Types::Hash.with_type_transform { |t| t.meta(omittable: true) }
schema = optional_keys.schema(name: 'strict.string', age: 'strict.int')
schema.(name: 'Jane') # => { name: 'Jane' }
```

From my perspective, these three methods is more than enough.

### DIFF
--- a/lib/dry/types/hash.rb
+++ b/lib/dry/types/hash.rb
@@ -9,17 +9,7 @@ module Dry
       # @param [Symbol] constructor
       # @return [Schema]
       def schema(type_map, constructor = nil)
-        type_fn = meta.fetch(:type_transform_fn, Schema::NO_TRANSFORM)
-        type_transform = Dry::Types::FnContainer[type_fn]
-
-        member_types = type_map.each_with_object({}) { |(name, type), result|
-          t = case type
-              when String, Class then Types[type]
-              else type
-              end
-
-          result[name] = type_transform.(t)
-        }
+        member_types = transform_types(type_map)
 
         if constructor.nil?
           Schema.new(primitive, member_types: member_types, **options, meta: meta)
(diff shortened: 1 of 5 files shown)


flash-gordon:
I discovered that `Schema#schema` that comes from `Hash` doesn't work as expected so I changed it to merge in input types.

```ruby
name_only = Types::Hash.schema(name: "strict.string")
with_email = name_only.schema(email: "string.string")
```

IMO this is more useful/less surprising.

flash-gordon:
I replaced `Schema#permissive` with `Schema#strict`, clearly, schemas should be permissive by default. Thanks @solnic ❤️

GustavoCaso:
@flash-gordon this is an amazing work ❤️. Really understandable

fledman:
the ability to set `omittable` on a per member basis is nice:
```ruby
normal = Types::Hash.schema({
  a: Types::Strict::String,
  b: Types::Strict::String.meta(omittable: true)
}, :strict)
normal.(a: 'A', b: 'B') # => {:a=>"A", :b=>"B"}
normal.(a: 'A') # => {:a=>"A"}
normal.(b: 'B') # Dry::Types::MissingKeyError: :a is missing in Hash input
```
I did notice that this doesn't work properly in reverse:
```ruby
reverse = Types::Hash.schema({
  a: Types::Strict::String.meta(omittable: false),
  b: Types::Strict::String
}, :schema)
reverse.(a: 'A', b: 'B') # => {:a=>"A", :b=>"B"}
reverse.(a: 'A') # => {:a=>"A"}
reverse.(b: 'B') # => {:b=>"B"}
```
because `omittable` is being overwritten in https://github.com/dry-rb/dry-types/blob/fb19568f44c9086b6761c6e6f77be3021555d988/lib/dry/types/hash/schema_builder.rb#L50
also it would be nice to have a shortcut so you can just write
```ruby
Types::Strict::String.omittable
```

flash-gordon:
@fledman old hash schemas are legacy, we could just remove them but since having them for backward compatibility is not a problem atm we'll keep them for a while, I'll just add a deprecation warning. If you're going to use `omittable` directly then use `Hash#schema` with a single argument. What comes to a helper method for `omittable` we could have it but I don't like the idea of bloating type's API (one of the reasons: it requires syncing the interface with `dry-struct`), will see how it goes.

PR #232, opened 2018-01-12
Rename int to integer

GustavoCaso:
@flash-gordon @solnic I renamed `int` to `integer`. Hope this helps


GustavoCaso:
No freaking way 😂 I will never commit after 12pm

GustavoCaso:
@solnic let's all forget this brain 💩 I had last night, and carry on with our life 😂 😂

PR #236, opened 2018-02-01
Include compat files

GustavoCaso:
@flash-gordon I was doing some experimentation and notice the compatibility files were not included.

I guess you planned to add when we reach `1.0`?


flash-gordon:
I didn’t, it’s on the user to load them. `compat.rb` is fine though, pls remove autoloading and merge it

Issue #213, opened 2017-10-10
[Dry::Types::Hash] No key provided when coercion fails


stephenminded:
If you fail to provide a value of the correct type when using a `Strict` type, you will get a `SchemaError` informing you what key and value violated the type. When using a `Coercible` type, a value that cannot be coerced just raises a `TypeError` with no extra information. It would be nice to get a similar `SchemaError` in this case so I could know which key and value caused the issue. E.g.

```ruby
require 'dry-types'
module Types
  include Dry::Types.module
end

hash = Types::Hash.schema(name: Types::Strict::String, age: Types::Coercible::Int)

hash[name: {}, age: '13']
# => Dry::Types::SchemaError: {} (Hash) has invalid type for :name violates constraints (type?(String, {}) failed)

hash[name: "Test", age: {}]
# => TypeError: can't convert Hash into Integer
```

stephenminded:
FWIW, I do have a branch which will raise a `SchemaError` when coercion fails. If you'd like me to open a PR I can do that https://github.com/dry-rb/dry-types/compare/master...stephenminded:provide_key_on_coercion_failure?expand=1

flash-gordon:
I included your commit to master, thanks 👍

PR #237, opened 2018-02-02
Types::Hash: Indicate which key could not be coerced properly

flash-gordon:
When using a coercible type and coercion fails, a `SchemaError` will now be raised and will indicate which key could not be coerced.


flash-gordon:
see #213

Issue #113, opened 2016-07-13
Feature request: hash schema with any number of (homogenous) members


StefanWiechula:
I'd like to validate a data structure that contains hashes where the keys and values each conform to schemas already expressible in dry-types but where the number of key/value pairs is not enumerable at type definition time.

Specifically, the use case I have now is a hash like:

```
{
42=>{:foo=>"aaa", :bar=>"bbb"},
99=>{:foo=>"ccc", :bar=>"ddd"},
...
33=>{:foo=>"xxx", :bar=>"yyy"},
}
```

Where the keys are all integers, necessarily unique (it's a Ruby Hash after all), and the values are hashes conforming to some schema already expressibly with dry-types.  E.g., following the example above:

```
Types::Hash.strict(
  foo: Types::Strict::String,
  bar: Types::Strict::String,
)
```

Hash schemas ( http://dry-rb.org/gems/dry-types/hash-schemas/ ) don't support this use case.  There is no way to express the outer collection structure without including all possible keys (infeasible in my application, they're arbitrary Ruby integers).

Perhaps something akin to array with member ( http://dry-rb.org/gems/dry-types/array-with-member/ ) could be added for hashes?  It would take a key type and a value type, then validate all keys and values against those when validating the parent hash.

Draiken:
I was in the middle of removing Virtus in favor of dry-types, and hit this issue.
Even worse, I had some other weird problems when calling `Thing.new` having different default values over time... Probably unrelated, but I had to ditch everything and either keep Virtus or write my own thing.

Wasted hours in what should have been a straight forward change :(

solnic:
@Draiken we don't have APIs that are matching Virtus features 100%. There's more work going into Struct API in a new project called [dry-struct](https://github.com/dry-rb/dry-struct) where related concerns will be handled. For now just stick to Virtus, although keep in mind that this may not be a straight-forward change as dry-types/struct are not a drop-in replacement for Virtus.

solnic:
How about an interface like that: `Types::Hash.keys(Types::Strict::Int)`?

Draiken:
That would cover it partially. Can I specify the type of the values the int
keys hold?
If yes, then it would be perfect for my use case

solnic:
Yes we could support that

mainameiz:
How about new type called `Map`? It is similar to `Hash` but will allow different number of keys of particular type (e.g. `String`).

flash-gordon:
[...]

PR #242, opened 2018-03-01
implement a Map type

fledman:
@flash-gordon 
implement a Map type based on the discussion from https://github.com/dry-rb/dry-types/issues/113

create one like:
```ruby
Dry::Types::Map.new(key_type: Dry::Types['strict.integer'], value_type: Dry::Types['strict.string'])
```
the empty<sup>[1]</sup> map is registered as `Dry::Types['map']` so you could also do:
```ruby
Dry::Types['map'].with(key_type: Dry::Types['strict.integer'], value_type: Dry::Types['strict.string'])
```
I chose to inherit from `Definition` as opposed to `Hash` to  avoid unintended interactions with the `Schema` logic

[1] by empty I mean unconstrained; any key-value pair is valid

_Edited to reflect the options renaming_

### DIFF
--- a/lib/dry/types/builder_methods.rb
+++ b/lib/dry/types/builder_methods.rb
@@ -83,7 +83,7 @@ module Dry
         Definition.new(klass).constructor(cons || block || klass.method(:new))
       end
 
-      # Build a definiton type
+      # Build a definition type
       #
       # @param [Class] klass
       #
@@ -91,6 +91,19 @@ module Dry
       # @api public
       def Definition(klass)
         Definition.new(klass)
+      end
+
+      # Build a map type
+      #
+      # @example
+      #   Types::IntMap = Types.Map(keys: Types::Integer)
+      #
+      # @param [Hash] options
+      #
+      # @return [Dry::Types::Map]
+      # @api public
+      def Map(**options)
+        Dry::Types::Map.new(**options)
       end
     end
   end
--- a/lib/dry/types/map.rb
+++ b/lib/dry/types/map.rb
@@ -1,0 +1,102 @@
+module Dry
+  module Types
+    class Map < Definition
+
+      def initialize(key_type: Any, value_type: Any, meta: EMPTY_HASH)
+        super(::Hash, key_type: key_type, value_type: value_type, meta: meta)
+        validate_options!
+      end
+
+      # @return [Type]
+      def key_type
+        options[:key_type]
+      end
+
+      # @return [Type]
+      def value_type
+        options[:value_type]
+      end
+
+      # @return [String]
+      def name
+        "Map"
+      end
+
+      # @param [Hash] hash
+      # @return [Hash]
+      def call(hash)
+        try(hash) do |failure|
+          raise MapError, failure.error.join("\n")
+        end.input
+      end
+      alias_method :[], :call
+
+      # @param [Hash] hash
+      # @return [Boolean]
+      def valid?(hash)
+        coerce(hash).success?
+      end
+      alias_method :===, :valid?
+
+      # @param [Hash] hash
+      # @return [Result]
+      def try(hash)
+        result = coerce(hash)
+        return result if result.success? || !block_given?
+        yield(result)
+      end
+
+      # @param meta [Boolean] Whether to dump the meta to the AST
+      # @return [Array] An AST representation
+      def to_ast(meta: true)
+        recurse = options.each_with_object({}) do |(k,v),h|
+                    h[k] = v.to_ast(meta: meta)
+                  end
+        [:map, [recurse, meta ? self.meta : EMPTY_HASH]]
+      end
+
+      # @param [Hash] new_options
+      # @return [Map]
+      def with(new_options)
+        self.class.new(**options, meta: @meta, **new_options)
+      end
+
+      private
+
+      def coerce(input)
+        return failure(
+          input, "#{input.inspect} must be an instance of #{primitive}"
+        ) unless primitive?(input)
+
+        output, failures = {}, []
+
+        input.each do |k,v|
+          res_k = options[:key_type].try(k)
+          res_v = options[:value_type].try(v)
+          if res_k.failure?
+            failures << "input key #{k.inspect} is invalid: #{res_k.error}"
+          elsif output.key?(res_k.input)
+            failures << "duplicate coerced hash key #{res_k.input.inspect}"
+          elsif res_v.failure?
+            failures << "input value #{v.inspect} for key #{k.inspect} is invalid: #{res_v.error}"
+          else
+            output[res_k.input] = res_v.input
+          end
+        end
+
+        return success(output) if failures.empty?
+
+        failure(input, failures)
+      end
+
+      def validate_options!
+        [:key_type, :value_type].each do |opt|
+          type = options[opt]
+          next if type.is_a?(Type)
+          raise MapError, ":#{opt} must be a #{Type}, got: #{type.inspect}"
+        end
+      end
+
+    end
+  end
+end
--- a/lib/dry/types/options.rb
+++ b/lib/dry/types/options.rb
@@ -6,9 +6,9 @@ module Dry
 
       # @see Definition#initialize
       def initialize(*args, meta: EMPTY_HASH, **options)
-        @__args__ = args
-        @options = options
-        @meta = meta
+        @__args__ = args.freeze
+        @options = options.freeze
+        @meta = meta.freeze
       end
 
       # @param [Hash] new_options
--- a/lib/dry/types/compiler.rb
+++ b/lib/dry/types/compiler.rb
@@ -98,6 +98,14 @@ module Dry
         Enum.new(visit(type), mapping: mapping, meta: meta)
       end
 
+      def visit_map(node)
+        options, meta = node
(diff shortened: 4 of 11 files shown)


lib/dry/types/builder_methods.rb
```diff
@@ -83,7 +83,7 @@ def Constructor(klass, cons = nil, &block)
         Definition.new(klass).constructor(cons || block || klass.method(:new))
       end
```
GustavoCaso: We probably going to need a builder method for building `Map`

```ruby
def Map(options={})
  Map.new(options)
end
```

lib/dry/types/map.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      # @return [String]
+      def name
+        "Map"
+      end
+
+      # @param [Hash] hash
+      # @return [Hash]
+      def call(hash)
+        try(hash) do |failure|
```
GustavoCaso: Following how `Array` and `Hash` are designed:
- the method `call` simple call the `coerce` which return the output or an Error
- the method `try` would also the `coerce` method and depending on the result it should return `Result {Success| Failure }` object and only yield the `Result::Failure` if a block was given

Maybe we can stick to that convention? WDYT? @fledman

fledman: So in my implementation, `coerce` is more similar to `Schema#resolve`, in that it always returns a `Result` object.

Then `call`/`try`/`valid?` build off of that to behave the same way as the convention:
  * call: return the output or raise an error
  * try:  yield or return the `Result` object
  * valid?: result.success?

I organized the code this way to cleanly separate the actual logic from error handling + output formatting.

fledman: More specifically, `Schema` implements `try` using `call`, while `Map` implements `call` using `try`. 

This reversal allows us avoid using raise/rescue for internal error handling e.g. https://github.com/dry-rb/dry-types/blob/f3c2561dbad3b1d44b564951c823983fdf30493d/lib/dry/types/hash/schema.rb#L56-L66

lib/dry/types/builder_methods.rb
fledman: seems reasonable, I will go ahead and add it

fledman: done

fledman:
@solnic also

lib/dry/types/map.rb
GustavoCaso: @fledman Thanks for the clarification. I'm not having a strong opinion about this, what I'm saying I'm ok with both solutions. I have created a gist https://gist.github.com/GustavoCaso/5376d5a0b5c195ec05c29464868543c2 just you can see the result, also I think the method `try` with this solution is more readable 😄.

Once again thank you for working on this ❤️

fledman: I generally prefer pushing raised errors to the edge of library code, and using other mechanisms  for internal handling.

But I'm not the owner here, so it's up to you.

lib/dry/types/map.rb
```diff
@@ -0,0 +1,110 @@
+module Dry
+  module Types
+    class Map < Definition
+
+      def initialize(keys: Undefined, values: Undefined, meta: EMPTY_HASH)
```
GustavoCaso: Hi @fledman one more question, hope you do not mind, why we are using `Undefined` for default values for `keys` and `values`? Maybe we could use `Dry::Types::Any` and we avoid all the checking of `Undefined` in the class.

Also, I think it would be better for the `ast` having an empty map the result would be more idiomatic:
```ruby
[:map, [{:keys=>[:definition, [Object, {}]], :values=>[:definition, [Object, {}]]}, {}]])
```

instead of:

```ruby
[:map, [{}, {}]]
```

WDYT?

lib/dry/types/map.rb
```diff
@@ -0,0 +1,110 @@
...
+        yield(result)
+      end
+
+      # @param meta [Boolean] Whether to dump the meta to the AST
+      # @return [Array] An AST representation
+      def to_ast(meta: true)
+        set = options.reject{ |_,v| Undefined.eql?(v) }
+        set.keys.each do |key|
+          set[key] = set[key].to_ast(meta: meta)
+        end
```
flash-gordon: This is too abstract for a very specific task: dump `key` and `value` option. Please remove that `Undefined` check and the loop, use `Any` instead of `Undefined`

lib/dry/types/map.rb
```diff
@@ -0,0 +1,110 @@
...
+      def if_undefined(receiver)
+        Undefined.eql?(receiver) ? yield : receiver
+      end
+
+      def coerce(input)
+        return failure(
+          input, "#{input.inspect} must be an instance of #{primitive}"
+        ) unless primitive?(input)
+
+        key_type = if_undefined(keys){ Any  }
```
flash-gordon: With `Any` as a default these two lines shouldn't be needed.

And JFYI I added `Undefined.default` recently: https://github.com/dry-rb/dry-core/blob/master/CHANGELOG.md#added

lib/dry/types/options.rb
```diff
@@ -6,9 +6,9 @@ module Options
 
       # @see Definition#initialize
       def initialize(*args, meta: EMPTY_HASH, **options)
-        @__args__ = args
-        @options = options
-        @meta = meta
+        @__args__ = args.freeze
```
flash-gordon: Please don't do this, dry-types doesn't offer an interface for mutation, this style of defensive programming doesn't really help, quite the opposite, it pollutes code and slows things down.

lib/dry/types/map.rb
```diff
@@ -0,0 +1,110 @@
...
+        options[:keys]
+      end
+
+      # @return [Type, Undefined]
+      def values
+        options[:values]
+      end
+
+      # @return [String]
+      def name
```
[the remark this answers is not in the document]
fledman: @flash-gordon 
are you sure? it seems to be defined for all of the built in types:
https://github.com/dry-rb/dry-types/blob/f3c2561dbad3b1d44b564951c823983fdf30493d/lib/dry/types/definition.rb#L39-L42
https://github.com/dry-rb/dry-types/blob/f3c2561dbad3b1d44b564951c823983fdf30493d/lib/dry/types/any.rb#L8-L11
https://github.com/dry-rb/dry-types/blob/f3c2561dbad3b1d44b564951c823983fdf30493d/lib/dry/types/sum.rb#L48-L51
https://github.com/dry-rb/dry-types/blob/f3c2561dbad3b1d44b564951c823983fdf30493d/lib/dry/types/constructor.rb#L42-L45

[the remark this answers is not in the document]
flash-gordon: @fledman doh! I've never seen it. Ok, this is fine but our introspection still sucks :)

lib/dry/types/map.rb
GustavoCaso: Once set `Any` as default this will be simplified

lib/dry/types/map.rb
```diff
@@ -0,0 +1,104 @@
+module Dry
+  module Types
+    class Map < Definition
+
+      def initialize(keys: Any, values: Any, meta: EMPTY_HASH)
```
fledman: @flash-gordon @GustavoCaso 
regarding naming:

`Map` is a top level type like `Integer`, `String`, `Array`, `Enum`, ... they are all singular.

But it doesn't seem like you have a strong pattern for plural type-init args.

e.g. array takes the singular `member`, but enum takes the plural `values`.

Personally I am not a huge fan of `keys` and `values`, but I think the singular `key` and `value` is even more confusing.

Do you have any other ideas? Maybe `key_type` and `value_type`?

lib/dry/types/map.rb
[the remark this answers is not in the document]
solnic: yeah it was just me doing whatever just to get some basic inspection in place

lib/dry/types/options.rb
fledman: @flash-gordon 

So the `@options` and `@meta` hashes are directly exposed, and can be freely mutated by any caller of `#options` and `#meta`; to preserve immutability, those two need to be frozen.

`@ __args__` is not exposed, but it seemed strange to not freeze it when I was freezing the other two. Also, I don't think it would slow performance; if anything, it would allow the interpreter to make better memory allocation decisions knowing that it won't be mutated.

lib/dry/types/map.rb
GustavoCaso: @fledman I also think that `key` and `value` could be confusing. I'm not against having `key_type` and `value_type`.

flash-gordon: `key_type`/`value_type` is fine

lib/dry/types/options.rb
flash-gordon: I won't argue on this because it's bikeshedding. But I don't think the interpreter would work faster, there are potential memory gains though but again they can be achieved with a generational GC (and cruby already has it).

fledman:
I've made changes; can you re-review?

Also, do you want me to squash this branch down to a single commit, or do you prefer to keep the entire history?

flash-gordon:
Thanks, I will take another look tomorrow most likely. You can squash this in the meantime, we don't have strict rules for it but squashing helps writing changelogs later so if you don't mind please do.

PR #250, opened 2018-05-04
Fix Default to allow accepting meta attributes

GustavoCaso:
We had a strange behavior combining `default` and `meta`

```ruby
type = Dry::Types['hash'].default({}).meta(omittable: true)
type[] # => {:meta=>{:omittable=>true}}
```

With these changes works as expected 

```ruby
type = Dry::Types['hash'].default({}).meta(omittable: true)
type[] # => { }
```


GustavoCaso:
@flash-gordon any update on this one?

PR #259, opened 2018-06-29
do not return default value if nil is passed

GustavoCaso:
@flash-gordon 

The issue was found in https://github.com/dry-rb/dry-struct/issues/90

Do not return the default value when passing `nil`

### DIFF
--- a/lib/dry/types/default.rb
+++ b/lib/dry/types/default.rb
@@ -67,7 +67,7 @@ module Dry
         if input.equal?(Undefined)
           evaluate
         else
-          type[input]
+          Undefined.default(type[input]) { evaluate }
         end
       end
       alias_method :[], :call
--- a/lib/dry/types/default.rb
+++ b/lib/dry/types/default.rb
@@ -67,8 +67,7 @@ module Dry
         if input.equal?(Undefined)
           evaluate
         else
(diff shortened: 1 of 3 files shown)


GustavoCaso:
@flash-gordon any update on this one?

flash-gordon:
The specs you modified ensured some specific behavior so I'm afraid we can't change it. Even though I don't like the idea of default values evaluated when the underlying type returns `nil` there is not much we can do about it besides amending the docs for now :(

macosgrove:
@flash-gordon Are you saying this can't be merged? I don't think it will be possible to replicate the existing behaviour of dry-types v0.12's `constructor_type :strict_with_defaults` without it, will it?

timriley:
@flash-gordon Is this the sort of thing we'd want to tee up for a breaking release (coordinated with any other of our gems that rely on this behaviour? I'd imagine dry-v with its params schemas would be one of them).

flash-gordon:
Since dry-validation is being re-written it's good timing to merge this in. I also made defaults to evaluate if the decorated type returns `Undefined`. Overall, this is a breaking change but I hope it'll be easy to upgrade. I wonder if we need a `{'',nil} => Undefined` type pre-defined

Issue #258, opened 2018-05-30
Cannot set constuctor for not constrained coercible type


ojab:
```ruby
float = Dry::Types['coercible.float'].constrained(gteq: 0).constructor do |input|
   input == :x ? 1.0 : input
end

float2 = Dry::Types['coercible.float'].constructor do |input|
  input == :x ? 1.0 : input
end

float[:x] # => 1.0
float2[:x] # => `Float': can't convert Symbol into Float (TypeError)
```

abrthel:
I have also ran into this issue but with sum types and constructor.

solnic:
Starting from `0.15.0` this can be handled via `Constructor#prepend` (there's `Constructor#append` too), ie:

```ruby
require 'dry/types'

module Types
  include Dry::Types.module
end

float = Types::Coercible::Float.constrained(gteq: 0).prepend { |input| input == :x ? 1.0 : input }

puts float[:x].inspect
# 1.0

puts float['1.0'].inspect
# 1.0

puts float['2.4'].inspect
# 2.4
```

PR #275, opened 2018-10-06
Add Constructor#{prepend,append}

flash-gordon:
Consequent calls of `.constructor` lead to right-to-left composition i.e. the block you're passing will be called _after_ the underlying constructor. Sometimes you want to prepend a block, that's what `Constructor#prepend` does. I also added Constructor#append as an alias for `#constructor`.

I believe this addresses #258

@solnic @timriley are you happy with names? E.g. it can be before/after

### DIFF
--- a/lib/dry/types/constructor.rb
+++ b/lib/dry/types/constructor.rb
@@ -44,6 +44,7 @@ module Dry
         type.name
       end
 
+      # @return [Boolean]
       def default?
         type.default?
       end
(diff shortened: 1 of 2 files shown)


timriley:
Nice work, @flash-gordon. I'm happy with append/prepend as names.

I'm going to find some time this week to see if this helps the issues I was having with rom-sql types. Will let you know how it goes.

flash-gordon:
@timriley actually, I'll merge this now and rebase https://github.com/dry-rb/dry-types/pull/286 on top of this so you can test them together

PR #278, opened 2018-12-23
Add metadata uris

viraptor:
Add the changelog / bug / source uris to the metadata section for easier
access.


flash-gordon:
Thanks!

PR #286, opened 2019-02-13
Represent hash schema as a list of keys

flash-gordon:
This changes the representation of hash schema. Now schema is a list of keys, each key is a `Hash::Key` object containing type, key name and the `required` flag (as opposed to `omittable` we had before). `required` won't be stored in `meta`, that's way cleaner design, it should have been done like that from the very beginning.

TODO:

- [x] fix specs
- [x] update compiler
- [x] add some more specs or update existing (i.e. replace `omittable` with `required`)
- [x] check compatibility with other gems, dry-schema and rom(-sql) first of all


lib/dry/types/hash/key.rb
```diff
@@ -0,0 +1,60 @@
...
+    class Hash < Definition
+      class Key
+        include Type
+        include ::Dry::Equalizer(:name, :type, :options)
+        include Decorator
+        include Builder
+
+        attr_reader :name
+
+        def initialize(type, name, required: Undefined, **options)
```
timriley: nitpicking, but putting the name first in the params list would feel more natural to me

lib/dry/types/hash/schema.rb
```diff
@@ -15,8 +15,10 @@ class Schema < Hash
         NO_TRANSFORM = Dry::Types::FnContainer.register { |x| x }
         SYMBOLIZE_KEY = Dry::Types::FnContainer.register(:to_sym.to_proc)
 
-        # @return [Hash{Symbol => Definition}]
-        attr_reader :member_types
+        # @return [Array[Dry::Types::Hash::Key]]
+        attr_reader :keys
+
+        attr_reader :index
```
timriley: `key_name_map` (or similar) might be a clearer name here?

lib/dry/types/hash/schema.rb
```diff
@@ -84,7 +89,7 @@ def to_ast(meta: true)
           [
             :hash_schema,
             [
-              member_types.map { |name, member| [:member, [name, member.to_ast(meta: meta)]] },
+              keys.map { |key| key.to_ast(meta: meta) },
```
timriley: This is a nice improvement :)

lib/dry/types/hash/key.rb
```diff
@@ -0,0 +1,60 @@
...
+          required = Undefined.default(required) do
+            type.meta.fetch(:required) { !type.meta.fetch(:omittable, false) }
+          end
```
timriley: I see down at the bottom of the diff on this PR (in `#transform_key_type`) that you have:

```ruby
key.required(false) if omittable?(constructor)
```

Is your thinking (either as later work in this PR, or in another subsequent one) that you might provide another way to provide _individually_ omittable keys _without_ having to rely on a meta attribute?  IIRC that was part of the bug we discovered in rom where it was exporting read hash schema ASTs with the metas excluded.

lib/dry/types/hash/key.rb
flash-gordon: to me as well, but the super constructor requires it in this order and changing the order of arguments when calling `super` is even more unnatural :)

lib/dry/types/hash/key.rb
flash-gordon: this is part of the plan, but I don't have all pieces in mind yet. It seems reasonable to support at least this way of defining omittable keys
```ruby
Types::Hash.schema(
  required: Types::Bool,
  optional?: Types::Bool
)
```

Another possible API is building a key object out of any type:

```ruby
Types::Strict::String.key(:name, required: false) # => #<Types::Hash::Key>
```

This is flexible, but because of OOP it tangles things a bit. 
Now still, we're yet to figure out if anything of this can help with that ROM issue. For that we first need to extract its description from the gitter archives :)

lib/dry/types/hash/key.rb
timriley: Oh! I didn't realise this. Fair enough then :)

lib/dry/types/hash/key.rb
timriley: @flash-gordon you're both and persuasive, I'll try do that this weekend :P

timriley: Apart from that, good to hear that it's part of your plan to support a better first-class ways of declaring omittable keys.

CHANGELOG.md
```diff
@@ -1,3 +1,41 @@
...
+  `Dry::Types::Hash#with_type_transform` now yields a key type instead of type + name:
+  ```ruby
+  Dry::Types['stirct.hash'].with_type_transform { |key| key.name == :age ? key.required(false) : key }
+  ```
+
+## Added
+
+- Optional keys for schemas can be provided with ?-ending symbols (flash-gordon)
+  ```ruby
+  Dry::Types['stirct.hash'].schema(name: 'strict.string', age?: 'strict.integer')
```
timriley: @flash-gordon FYI, there's a `stirct` typo across all of these CHANGELOG entries that you'll want to fix :)

CHANGELOG.md
```diff
@@ -1,3 +1,41 @@
...
+  ```
+- Another way of making keys optional is setting `required: false` to meta. In fact, it is the preferrable
+  way if you have to store this information in `meta`, otherwise use the Key's API (see below) (flash-gordon)
```
timriley: @flash-gordon In reading this, what's not clear to me is _why_ this is preferable in this case. Is it because you're already adding to the meta so it's easier just to add that extra `required: false` entry at the same time? Is this _really_ preferable here, or is it more it's just somewhat more _convenient?_

CHANGELOG.md
```diff
@@ -1,3 +1,41 @@
...
+  Dry::Types['stirct.hash'].with_type_transform { |key| key.name == :age ? key.required(false) : key }
+  ```
+
+## Added
+
+- Optional keys for schemas can be provided with ?-ending symbols (flash-gordon)
+  ```ruby
+  Dry::Types['stirct.hash'].schema(name: 'strict.string', age?: 'strict.integer')
+  ```
+- Another way of making keys optional is setting `required: false` to meta. In fact, it is the preferrable
```
timriley: Typo: `preferrable` -> `preferable`

CHANGELOG.md
```diff
@@ -1,3 +1,41 @@
...
+
+- [BREAKING] Internal representation of hash schemas was changed to be a simple list of key types (flash-gordon)
+  `Dry::Types::Hash#with_type_transform` now yields a key type instead of type + name:
+  ```ruby
+  Dry::Types['stirct.hash'].with_type_transform { |key| key.name == :age ? key.required(false) : key }
+  ```
+
+## Added
+
+- Optional keys for schemas can be provided with ?-ending symbols (flash-gordon)
```
timriley: Do we want to put an explainer in here for how to use ?-ending names when we _don't_ want the implicit optional key behaviour?

CHANGELOG.md
flash-gordon: because consistency, we now use `required` under the hood and don't rely on fragile metadata so naming things straight is the preferable way. `:required` from `meta` goes to `attr_reader :required` of `Key`

CHANGELOG.md
flash-gordon: actually, I had no plans for that. I'd like to see a real use case for that first

PR #287, opened 2019-02-20
fix detecting invalid integers

exterm:
`Dry::Types['coercible.integer']` uses `Integer()` as a constructor, which raises `ArgumentError` if it can't coerce the input.

`Dry::Types::Constructor#valid?` only caught `TypeError` and `NoMethodError`, so it would raise for checking the validity of an invalid integer.


flash-gordon:
Thanks!

PR #295, opened 2019-02-27
Remove registration of custom types in specs

### LINKED ISSUE #288: Remove registration of custom types in specs
We don't expose `Dry::Types.register` as public API anymore, using it in specs looks confusing, we need to remove the usages.

skryukov:
Fixes #288


flash-gordon:
thank you!

PR #296, opened 2019-02-28
Print a warning when default value isn't frozen

### LINKED ISSUE #279: Print a warning when default value isn't frozen
This occasionally happens that people think a default value won't be shared across type applications. That's not the case. It could be avoided through the use of `.dup` on values but I argue it's not always safe, rather I'd print a warning that the provided value isn't frozen. Also, for cases when it's fine for some reason, I'd have a `shared` keyword that would discard the warning.
```ruby
Types::Strict::String.default('some') # => warning: 'some' is mutable
Types::Strict::String.default('some'.freeze) # => no warning
Types::Strict::String.default('some', shared: true) # => no warning
```

skryukov:
Fixes #279

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -32,10 +32,18 @@ module Dry
       end
 
       # @param [Object] input
+      # @param [Hash] options
       # @param [#call,nil] block
       # @raise [ConstraintError]
       # @return [Default]
-      def default(input = Undefined, &block)
+      def default(input = Undefined, options = EMPTY_HASH, &block)
+        unless input.frozen? || options[:shared]
+          Dry::Core::Deprecations.warn("#{input.inspect} is mutable."\
+            ' Be careful: types will return the same instance of the default'\
+            ' value every time. Call `.freeze` when setting the default'\
+            ' or pass `shared: true` to discard this warning.', tag: :'dry-types')
+        end
+
         value = input.equal?(Undefined) ? block : input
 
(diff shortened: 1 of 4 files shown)


lib/dry/types/builder.rb
```diff
@@ -32,10 +32,17 @@ def constrained(options)
...
 
       # @param [Object] input
+      # @param [Hash] options
       # @param [#call,nil] block
       # @raise [ConstraintError]
       # @return [Default]
-      def default(input = Undefined, &block)
+      def default(input = Undefined, options = {}, &block)
+        unless input.frozen? || options[:shared]
+          Kernel.warn "#{input.inspect} is mutable. Be careful: types will return "\
```
flash-gordon: Please use `Dry::Core::Deprecations` for that, it produces tagged messages

lib/dry/types/builder.rb
```diff
@@ -32,10 +32,17 @@ def constrained(options)
       end
 
       # @param [Object] input
+      # @param [Hash] options
       # @param [#call,nil] block
       # @raise [ConstraintError]
       # @return [Default]
-      def default(input = Undefined, &block)
+      def default(input = Undefined, options = {}, &block)
```
flash-gordon:
```suggestion
      def default(input = Undefined, options = EMPTY_HASH, &block)
```

skryukov:
@flash-gordon thanks for review! Force pushed fixed version.

Issue #290, opened 2019-02-24
Better types introspection


flash-gordon:
It's really easy to get lost with the output of an average-sized type.
Work on this is already in progress and I'll push some PoC shortly.
https://github.com/dry-rb/dry-equalizer/pull/13 is related

PR #299, opened 2019-03-08
Add type printer with better type introspection

flash-gordon:
Addresses #290 

This is a WIP. The goal is to print only what needs to be print. I disabled the default `inspect/to_s` implementation coming from dry-equalizer and replaced it with a custom traversing object.

### DIFF
--- a/spec/dry/types/constrained_spec.rb
+++ b/spec/dry/types/constrained_spec.rb
@@ -51,6 +51,20 @@ RSpec.describe Dry::Types::Constrained do
     end
   end
 
+  describe '#to_s' do
+    subject(:type) do
+      Dry::Types['strict.string'].constrained(size: 3..12)
+    end
+
+    it 'returns string representation of the type' do
+      expect(type.to_s).
+        to eql(
+            "#<Dry::Types[Constrained<Definition<String> "\
+            "rule=[type?(String) AND size?(3..12)]>]>"
+          )
+    end
+  end
+
   context 'with a constructor type' do
     subject(:type) do
       Dry::Types['coercible.hash'].constrained(size: 1)
--- a/spec/dry/types/default_spec.rb
+++ b/spec/dry/types/default_spec.rb
@@ -165,4 +165,80 @@ RSpec.describe Dry::Types::Definition, '#default' do
       end
     end
   end
+
+  describe '#to_s' do
+    context 'static valute' do
+      subject(:type) { Dry::Types['string'].default('foo') }
+
+      it 'returns string representation of the type' do
+        expect(type.to_s).to eql('#<Dry::Types[Default<Definition<String> value="foo">]>')
+      end
+    end
+
+    context 'callable value' do
+      subject(:type) { Dry::Types['string'].default(value_constructor) }
+
+      context 'proc' do
+        let(:value_constructor) { proc { 'foo' } }
+
+        let(:line_no) { value_constructor.source_location[1] }
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
+              "#<Dry::Types[Default<Definition<String> "\
+              "value_fn=spec/dry/types/default_spec.rb:#{ line_no }>]>"
+            )
+        end
+      end
+
+      context 'proc w/o source' do
+        let(:value_constructor) { method(:Integer).to_proc }
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
+              "#<Dry::Types[Default<Definition<String> "\
+              "value_fn=(lambda)>]>"
+            )
+        end
+      end
+
+      context 'method' do
+        let(:value_constructor) { Kernel.method(:Integer) }
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
+              "#<Dry::Types[Default<Definition<String> "\
+              "value_fn=Kernel.Integer>]>"
+            )
+        end
+      end
+
+      context 'callable object' do
+        let(:value_constructor) do
+          obj = Object.new
+
+          def obj.to_s
+            "callable"
+          end
+
+          def obj.call(*)
+            5
+          end
+
+          obj
+        end
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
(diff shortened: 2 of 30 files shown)


spec/dry/types/schema_spec.rb
```diff
@@ -315,4 +315,57 @@ def self.to_ary
...
+    context 'key transformation' do
+      let(:key_transformation) { :to_sym.to_proc }
+
+      let(:transformation_name) { Dry::Types::FnContainer.register_name(key_transformation) }
+
+      subject(:type) { Dry::Types['hash'].schema({}).with_key_transform(key_transformation) }
+
+      it 'returns string representation of the type' do
+        expect(type.to_s).
+          to eql("#<Dry::Types[Schema<keys={} key_fn=#{ transformation_name }.call>]>")
```
timriley: Is there a way we can specify the output explicitly here, without relying on `transformation_name.to_s` via this interpolation? Just reading through the specs as you suggested and this one made it harder for me to understand the final output.

spec/dry/types/default_spec.rb
```diff
@@ -165,4 +165,80 @@
...
+      context 'proc' do
+        let(:value_constructor) { proc { 'foo' } }
+
+        let(:line_no) { value_constructor.source_location[1] }
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
+              "#<Dry::Types[Default<Definition<String> "\
+              "value_fn=spec/dry/types/default_spec.rb:#{ line_no }>]>"
```
timriley: This is bloody spectacular!! 😄

spec/dry/types/default_spec.rb
```diff
@@ -165,4 +165,80 @@
...
+          end
+
+          obj
+        end
+
+        it 'returns string representation of the type' do
+          expect(type.to_s).
+            to eql(
+              "#<Dry::Types[Default<Definition<String> "\
+              "value_fn=callable.call>]>"
```
timriley: Would it make more sense for this to be `callable#call`?

spec/dry/types/constrained_spec.rb
```diff
@@ -51,6 +51,20 @@
...
+  describe '#to_s' do
+    subject(:type) do
+      Dry::Types['strict.string'].constrained(size: 3..12)
+    end
+
+    it 'returns string representation of the type' do
+      expect(type.to_s).
+        to eql(
+            "#<Dry::Types[Constrained<Definition<String> "\
+            "rule=[type?(String) AND size?(3..12)]>]>"
```
timriley: I love the " AND " here 👍

spec/dry/types/schema_spec.rb
flash-gordon: not really, its name is dynamic because functions/procs are registered globally in `Dry::Types::FnContainer`, it'll be something like `fn_70170547805360`

spec/dry/types/default_spec.rb
flash-gordon: technically, it's `.call`. Hashtag is for class methods, i.e. `SomeCallableClass#call`, it'd be wrong to do `"#{ callable.class }#call"` since we rely on the instance's interface

spec/dry/types/constrained_spec.rb
flash-gordon: Actually, this comes straight from dry-logic, I did nothing to represent it this :)

spec/dry/types/default_spec.rb
timriley: > Hashtag is for class methods

@flash-gordon are you _sure?_ For the longest time I've used `#` as the delimiter when denoting a class' _instance_ methods. As far as I can tell, this is in accordance with community convention, e.g. on https://docs.ruby-lang.org/en/2.6.0/String.html you'll see instance methods prefixed by `#`:

screen shot 2019-03-08 at 9 44 49 pm

spec/dry/types/schema_spec.rb
timriley: @flash-gordon Ah, fair enough.

spec/dry/types/default_spec.rb
timriley: oh wait, I think I understand what you're saying here: because we're not actually referring to the _class_, we can't use the class-based notation for its instance methods (e.g. `Class#method`)... is that right?

flash-gordon: @timriley sorry, I didn't make it clear. Here we rely on the fact that the object producing default values responds to `:call`. It doesn't mean it has the `call` method defined in its class, that was my point. However, now I think in this extreme case we can call `.method(:call)` (bad pun) and see where it's actually present. Then print different output depending on where its definition is.

timriley: @flash-gordon Cool, I fully understand now, thanks for the discussion :) And your idea to take this one step further sounds good, too. 👍

spec/dry/types/constrained_spec.rb
solnic: @timriley oh you didn't know that dry logic rules are coercible to nice string representation? Check this out:

``` ruby
schema = Dry::Schema.Params do
  required(:email).filled(:string)
  required(:age).filled(:integer, gt?: 18)
  required(:phone) { array? & min_size?(1) & each(:str?, min_size?: 6) | str? & min_size?(6) }
end

schema.rules.values.map(&:to_s)
# [
#   "key?(:email) AND key[email](str? AND filled?)",
#   "key?(:age) AND key[age](int? AND filled? AND gt?(18))", "key?(:phone) AND key[phone](array? AND min_size?(1) AND each(str? AND min_size?(6)) OR str? AND min_size?(6))"
# ]
```

flash-gordon:
I think I'll stop here for now. I experimented with different representations but wasn't satisfied with the results. I should have studied free monads 😟 One day I'll get back to this

flash-gordon:
Codeclimate complains after refactoring to continuation passing style though in my opinion, it's a clear win here.

spec/dry/types/constrained_spec.rb
timriley: @solnic that is super nice :)

Issue #292, opened 2019-02-24
Partial schema application


flash-gordon:
There are cases where we want to partial input. For example, we have a hash which conforms to a schema and we want to update one value in this hash. We need to merge it first and re-apply the whole schema once again, even if we know only one key is affected. This is what happens in `Dry::Struct#new`, partial schema application could speed this up. Note, this would improve ROM's performance as well.

flash-gordon:
implemented in #301, some improvements did later in master

PR #301, opened 2019-03-09
Add option for ignoring missing keys

flash-gordon:
It'll be useful for dry-struct where new struct can be created based on existing ones. Now we can check only new keys and skip type application for previously existed.


flash-gordon:
refs #292

PR #303, opened 2019-03-11
Options for Dry::Types.module

### LINKED ISSUE #289: Options for Dry::Types.module
We want to make dynamic module creation configurable so that the user can choose whether they want their top-level types to be strict/nominal/coercible/etc. Roughly speaking,

```ruby
module Types
  include Dry::Types.module(:strict)
end
```

makes `Types::Integer` to be strict.

We should have done this earlier, now it's scheduled for the 1.0 release.

flash-gordon:
Closes #289

### DIFF
--- a/lib/dry/types.rb
+++ b/lib/dry/types.rb
@@ -30,11 +30,11 @@ module Dry
     TYPE_SPEC_REGEX = %r[(.+)<(.+)>].freeze
 
     # @return [Module]
-    def self.module
-      namespace = Module.new
-      define_constants(namespace, type_keys)
-      namespace.extend(BuilderMethods)
-      namespace
+    def self.module(*namespaces)
+      mod = Module.new
+      define_constants(mod, type_keys(namespaces))
+      mod.extend(BuilderMethods)
+      mod
     end
 
     # @return [Container{String => Definition}]
@@ -91,7 +91,7 @@ module Dry
         [Inflector.camelize(parts.pop), parts.map(&Inflector.method(:camelize))]
       end
 
-      names.map do |(klass, parts)|
+      names.map do |klass, parts|
         mod = parts.reduce(namespace) do |a, e|
           a.constants.include?(e.to_sym) ? a.const_get(e) : a.const_set(e, Module.new)
         end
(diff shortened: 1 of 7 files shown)


flash-gordon:
@solnic @timriley let's discuss this, see examples in docs and specs. I'll rename `Definition` to `Nominal` separately

lib/dry/types.rb
```diff
@@ -27,29 +27,55 @@ module Types
...
+    #   # nominal types are exported by default
+    #   Types::Integer
+    #   # => #<Dry::Types[Definition<Integer>]>
+    #   Types::Strict::Integer
+    #   # => #<Dry::Types[Constrained<Definition<Integer> rule=[type?(Integer)]>]>
+    #
+    # @example changing default types
+    #
+    #   module Types
+    #     include Dry::Types(default: :strict)
```
solnic: There's no `Dry.Types` method, do you intend to add it eventually?

lib/dry/types.rb
```diff
@@ -126,19 +134,13 @@ def self.type_map
       @type_map ||= Concurrent::Map.new
     end
 
-    # List of type keys defined in {Dry::Types.container}
-    # @return [<String>]
-    def self.type_keys
-      container.keys
-    end
-
     private
```
solnic: This is not needed as it has no effect

lib/dry/types.rb
flash-gordon: yeah, will do it today and merge the whole thing shortly after that

lib/dry/types.rb
```diff
@@ -148,6 +110,55 @@ def self.const_missing(const)
...
     end
   end
+
+  # Export registered types as a module with constants
+  #
+  # @example no options
+  #
+  #   module Types
+  #     # exports all types as constants, uses modules for namespaces
+  #     include Dry::Types.module
```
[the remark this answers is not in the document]
flash-gordon: right! Thanks for catching it, no idea why I formulated it like this

lib/dry/types/module.rb
```diff
@@ -0,0 +1,103 @@
...
+          modules.reduce(tree) { |br, name| br[name] ||= {} }[const_name] = type
+        }.freeze
+      end
+
+      private
+
+      # @api private
+      def check_parameters(*namespaces, default: Undefined, **aliases)
+        referenced = namespaces.dup
+        referenced << default unless false.equal?(Undefined.default(default, false))
```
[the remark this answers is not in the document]
flash-gordon: nevermind, I tried to rewrite it several times but at the end of the day it's simply `false.equal?(default) || Undefined.equal?(default)`, I'll change it in master

flash-gordon:
@timriley wrt aliases I don't have an example right now, but since people import constants in _their_ code it makes sense to give them control over the process

PR #318, opened 2019-04-17
Split call into call_safe and call_unsafe

flash-gordon:
The idea is to choose the execution path in the top-level type. This PR also eliminates super calls in functions of constructor types, makes them faster for free.

```
$ bundle exec ruby benchmarks/schema_valid_vs_invalid.rb 
Warming up --------------------------------------
         valid input    84.216k i/100ms
       invalid input    85.912k i/100ms
Calculating -------------------------------------
         valid input      1.088M (± 3.7%) i/s -      5.474M in   5.037302s
       invalid input      1.098M (± 1.4%) i/s -      5.498M in   5.006690s

Comparison:
       invalid input:  1098412.8 i/s
         valid input:  1088334.7 i/s - same-ish: difference falls within error
```

### DIFF
--- a/lib/dry/types/array/member.rb
+++ b/lib/dry/types/array/member.rb
@@ -15,27 +15,31 @@ module Dry
           super
         end
 
-        # @param [Object] input
-        # @return [Array]
-        def call(input, &block)
+        def call_unsafe(input)
           if primitive?(input)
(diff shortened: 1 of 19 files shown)


flash-gordon:
Note that these methods are private and the whole thing is only about performance

PR #330, opened 2019-06-21
Add Coercible::Symbol, Params::Symbol & JSON::Symbol

### LINKED ISSUE #47: Add support for Symbol primitive

waiting-for-dev:
I have kept the distinction between kernel or by convention method coercion in the implementation, so it would be easy to separate them as different families if desired in the future.

Closes #47

### DIFF
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -88,6 +88,19 @@ module Dry
         end
       end
 
+      # @param [#to_sym, Object] input
+      #
+      # @return [Symbol, Object]
+      #
+      # @raise CoercionError
+      #
+      # @api public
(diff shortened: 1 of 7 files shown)


flash-gordon:
Thanks!
@solnic @timriley do you have any objections against having Symbol in params and json? I don't.

solnic:
> @solnic @timriley do you have any objections against having Symbol in params and json? I don't.

Yea it should be there too. I actually have a use-case for it already in my app heh.

flash-gordon:
ok, I'll merge it tomorrow and add some other stuff too, so we can release 1.1.0 next week

PR #331, opened 2019-06-22
Add a patch restoring old way of making schema keys omittable

flash-gordon:


solnic:
Maybe it should use a deprecation warning?

flash-gordon:
Actually, you're right.

PR #333, opened 2019-06-22
Add Array::Constructor

flash-gordon:
This fixes a few corner cases when using array types. One specific was reported by @solnic 

```ruby
Dry::Types['params.array<params.integer>'].member.('1')
# => Dry::Types::CoercionError: "1" cannot be coerced to array
# from /Users/gordon/dev/dry-rb/dry-types/lib/dry/types/coercions/params.rb:125:in `to_ary'
```

Now it works fine. This is the same solution we have for hash constructors so it potentially can be generalized but atm I just want to fix it as it is.

### DIFF
--- a/lib/dry/types/array.rb
+++ b/lib/dry/types/array.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'dry/types/array/member'
+require 'dry/types/array/constructor'
 
 module Dry
   module Types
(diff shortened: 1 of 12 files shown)


flash-gordon:
Well, turns out it doesn't really work (breaks dry-schema) and we need a generalized way of determining type composability, I'll be working on this.

PR #334, opened 2019-06-24
Rename Types.Contract to Types.Interface

flash-gordon:
@waiting-for-dev it's suggested by @solnic. His concern is people may confuse it with `Dry::Validation::Contract`. I don't have an opinion on the matter but if you do, please chime in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## Added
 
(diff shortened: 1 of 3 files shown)


flash-gordon:
also, we thought about naming it `Quack` or `Duck`. jk

waiting-for-dev:
It's ok, I don't mind too much about the name. Better use `Interface` to avoid confusions, then.

PR #345, opened 2019-08-06
Empty strings for integer params should be coerced to nil

### LINKED ISSUE #344: Types::Params::Integer fails on empty strings
Form parameters return an empty string for a `nil` integer, but dry-types `Types::Params::Integer` raises an error on this: `Dry::Types::CoercionError: invalid value for Integer(): ""`.

This should be returning `nil`, as I understand the difference between `Types::Params::Integer` and `Types::Coercible::Integer`.  I've monkey-patched the method to do so in my environment, and this is a simple change.  I can open a PR if this is desired.

CodingAnarchy:
Coerce empty string integer params to `nil`, to fix #344.

### DIFF
--- a/lib/dry/types/coercions/params.rb
+++ b/lib/dry/types/coercions/params.rb
@@ -66,7 +66,9 @@ module Dry
         #
         # @api public
         def self.to_int(input, &block)
-          if input.is_a? String
+          if empty_str?(input)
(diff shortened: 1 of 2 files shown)


flash-gordon:
Closed in favor of #348

PR #353, opened 2019-09-15
Add monads extension

### LINKED ISSUE #325: [Feature request] An extension that makes #try return a monad
It would be really nice if `Foo.try({..})` returned a `Dry::Monads::Try` or `Result`.

skryukov:
This extension adds `.to_monad` method which replaces `Dry::Types::Result` with `Dry::Monads::Result`:

```ruby
Dry::Types.load_extensions(:monads)

result = Dry::Types['strict.string'].try('Jane')
result.class #=> Dry::Types::Result::Success
monad = result.to_monad
monad.class #=> Dry::Monads::Result::Success
monad.value!  # => 'Jane'

result = Dry::Types['strict.string'].try(nil)
result.class #=> Dry::Types::Failure
monad = result.to_monad
monad.class #=> Dry::Monads::Result::Failure
monad.failure  # => [#<Dry::Types::ConstraintError>, nil]
```

Closes #325

### DIFF
--- a/lib/dry/types/extensions/monads.rb
+++ b/lib/dry/types/extensions/monads.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'dry/monads/result'
+
+module Dry
+  module Types
+    # Monad extension for Result
+    #
+    # @api public
+    class Result
+      include Dry::Monads::Result::Mixin
+
+      # Turn result into a monad
+      #
+      # This makes result objects work with dry-monads (or anything with a compatible interface)
+      #
+      # @return [Dry::Monads::Success,Dry::Monads::Failure]
+      #
+      # @api public
+      def to_monad
+        if success?
+          Success(input)
+        else
+          Failure([error, input])
+        end
+      end
+    end
+  end
+end
--- a/spec/extensions/monads/result_spec.rb
+++ b/spec/extensions/monads/result_spec.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+RSpec.describe Dry::Types::Result do
+  before { Dry::Types.load_extensions(:monads) }
+
+  let(:type) { Dry::Types['strict.string'] }
+
+  let(:result) { type.try(input) }
+
+  context 'interface' do
+    let(:input) { {} }
+
+    it 'responds to #to_monad' do
+      expect(result).to respond_to(:to_monad)
+    end
+  end
+
+  context 'with valid input' do
+    let(:input) { 'Jane' }
+
+    describe '#to_monad' do
+      it 'wraps Result with Success' do
+        monad = result.to_monad
+
+        expect(monad).to be_a Dry::Monads::Result
+        expect(monad).to be_success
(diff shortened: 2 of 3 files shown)


lib/dry/types/extensions/monads.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Turn result into a monad
+      #
+      # This makes result objects work with dry-monads (or anything with a compatible interface)
+      #
+      # @return [Dry::Monads::Success,Dry::Monads::Failure]
+      #
+      # @api public
+      def to_monad
+        if success?
+          Success(self)
```
beezee: How do you feel about folding the Dry::Types::Result into the Dry::Monads::Result instead of nesting?

It would make Dry::Monads::Success<Dry::Types::Failure<x>> and Dry::Monads::Failure<Dry::Types::Success<x>> cases that a user doesn't have to think about for exhaustivity despite being impossible via this code path, and also just removes a level of nesting.

spec/extensions/monads/result_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  context 'with valid input' do
+    let(:input) { 'Jane' }
+
+    describe '#to_monad' do
+      it 'wraps Result with Success' do
+        monad = result.to_monad
+
+        expect(monad).to be_a Dry::Monads::Result
+        expect(monad).to be_success
+        expect(monad.value!).to be(result)
```
beezee: Per my comment re: nesting, this could be `expect(monad.value!).to be(input)`

lib/dry/types/extensions/monads.rb
beezee: though I see the nesting is exactly how it's done in schema, so I suppose my proposal is larger in scope than I intended.

spec/extensions/monads/result_spec.rb
skryukov: > Per my comment re: nesting, this could be `expect(monad.value!).to be(input)`

Oh, ok, I will do it =)

beezee: i wonder what @flash-gordon thinks about the idea since it deviates from the example he pointed out in schema does. imo this would be nicer behavior across the board but i recognize that's a much bigger change and modifies existing behavior and needs more thought whether to be a breaking change vs a new method with modified behavior.

skryukov: I think we can use `.to_monad` method and wrap error/input with `Dry::Monads::Result`:
```ruby
result = Dry::Types['strict.string'].try('Jane')
result.class #=> Dry::Types::Success
monad = result.to_monad
monad.class #=> Dry::Monads::Success
result.value == monad.value!  # => true

result = Dry::Types['strict.string'].try(nil)
result.class #=> Dry::Types::Failure
monad = result.to_monad
monad.class #=> Dry::Monads::Failure
result.error == monad.failure  # => true
```

Still this can be problematic... For example in case of extension of `Dry::Types::Result` with some metadata in the feature.

flash-gordon:
Thanks!

PR #366, opened 2019-10-22
Fix error on Dry::Types::Array#try

esparta:
One of the branches was calling `Nominal#failure` with a wrong parameter, and that was causing an exception that should not be happening.

Also adding more spec for `Dry::Types::Array#try`


flash-gordon:
thank you!

esparta:
:) no problemo. Thank you for hard work on the dry-* gems :+1:

PR #401, opened 2020-05-14
Document use of custom method in Constructor

timstott:
Add example in `Types.Constructor` on how to use a custom method.

The documentation will now cover the signature of `Types.Constructor(klass, cons = nil, &block)` in full.

### DIFF
--- a/docsite/source/custom-types.html.md
+++ b/docsite/source/custom-types.html.md
@@ -47,6 +47,9 @@ user_type = Types.Constructor(User)
 # It is equivalent to User.new(name: 'John')
 user_type[name: 'John']
 
+# Using a method User.build
+user_type = Types.Constructor(User, &:build)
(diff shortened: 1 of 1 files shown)


flash-gordon:
Hm, it doesn't look right
```ruby
class User
  def self.build(values)
    new(values)
  end

  def initialize(attributes)
  end
end

UserType = Types.Constructor(User, &:build)

UserType.(name: 'John') # => Dry::Types::CoercionError: undefined method `build' for {:name=>"John"}:Hash
```

@timstott what did you try to explain with that example?

timstott:
You are absolutely correct @flash-gordon, it should be

```ruby
UserType = Types.Constructor(User, User.method(:build))
```

I pushed a fix in #402, and apologize for the mix-up.

PR #402, opened 2020-05-16
Fix Constructor custom method documentation

timstott:
Fix the documentation bug I introduced in dry-rb/dry-types#401


flash-gordon:
thanks!

PR #406, opened 2020-10-08
Cherry-pick “concurrent/map” require

timriley:
We don’t need the whole of the concurrent gem, just this one part of it.

This results in 242 fewer files loaded when requiring “dry/types”.


flash-gordon (approved):

solnic (approved):
🙇

PR #409, opened 2020-12-14
Add `#or_nil`

ianks:
So there have been a few times where it would be nice to simply nullify a value if a type is invalid. Imagine we have this type:

```ruby
EmailAddress ||= begin
  Types::String
    .constructor(&:downcase)
    .constructor(&:strip)
    .constrained(format: URI::MailTo::EMAIL_REGEXP)
end
```

Currently, you can use `.lax` to rescue and return the input: `EmailAddress.lax.call('bad') #=> 'bad'`

This is useful, but it would useful to be able to nil out the value, as well. This would make it much easier to filter our junky data that’s is not business critical.

This PR adds `#or_nil` to support this exact use case

```ruby
EmailAddress.or_nil.call('test@test.com') #=> 'test@test.com'`
EmailAddress.or_nil.call('bad') #=> nil`
```

via: https://discourse.dry-rb.org/t/something-like-lax-but-return-nil-instead-of-input/1168/2

### DIFF
--- a/changelog.yml
+++ b/changelog.yml
@@ -1,4 +1,8 @@
 ---
+- version: 1.5.0
+  date: '2020-12-14'
+  added:
+  - Add `#or_nil` to support nullifying invalid values (@ianks)
 - version: 1.4.0
   date: '2020-03-09'
   fixed:
--- a/changelog.yml
+++ b/changelog.yml
@@ -1,8 +1,4 @@
 ---
-- version: 1.5.0
-  date: '2020-12-14'
-  added:
-  - Add `#or_nil` to support nullifying invalid values (@ianks)
 - version: 1.4.0
   date: '2020-03-09'
   fixed:
--- a/lib/dry/types/or_nil.rb
+++ b/lib/dry/types/or_nil.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'dry/types/decorator'
+require "dry/types/decorator"
 
 module Dry
   module Types
--- a/lib/dry/types/or_nil.rb
+++ b/lib/dry/types/or_nil.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'dry/types/decorator'
+
+module Dry
+  module Types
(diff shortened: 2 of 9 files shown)


changelog.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- version: 1.5.0
+  date: '2020-12-14'
+  added:
+  - Add `#or_nil` to support nullifying invalid values (@ianks)
```
solnic, line 5: @ianks please remove this part because we update changelogs via PR merge commit messages, and yes, this should be documented in that file and contribution guidelines, which I'll do shortly 🙂

lib/dry/types/or_nil.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require "dry/types/decorator"
+
+module Dry
+  module Types
+    # Lax types rescue from type-related errors when constructors fail and return nil
```
solnic, line 7: Shouldn't this say `OrNil types ...`?

solnic (changes requested):
Just a couple of minor things to address, otherwise LGTM! 🙇🏻

solnic:
@ianks I'm *so sorry* but I didn't see @flash-gordon's reply right here: https://discourse.dry-rb.org/t/something-like-lax-but-return-nil-instead-of-input/1168/5 before I suggested `OrNil` could be a good addition. Could you please sync up with Nikita about it? From what I *now* know, he started working on something that will simply support what *you need* too, and will be more generic.

ianks:
Would it be possible to merge this PR @flash-gordon / @solnic? Then the refactoring can make this feature work with it? This functionality is critical for us and wouldn't want to have to delay it a lot if that can be avoided.

flash-gordon:
@ianks can you pin your dependency to this branch maybe?

flash-gordon:
Now in master it can be done with
```ruby
Dry::Types.define_builder(:or_nil) { _1.optional.fallback(nil) }
```
🎉

Issue #407, opened 2020-10-29
Type#fallback to mock previous implementation of Type#default


pyromaniac:
Hey, thanks for a great gems first of all.

I was migrating from dry-types `0.14.0` and faced with the change of `default` behavior. There was even a type transformation proposed for dry-struct but I thought that maybe we can address such core functionality on a lower level.

I did the following patch:

```ruby
Dry::Types::Builder.class_eval do
  # A ripoff of a `default` implementation,
  # except it is applied after coercion
  # and doesn't support blocks.
  def fallback(input)
    unless input.frozen?
      where = Dry::Core::Deprecations::STACK.()
      Dry::Core::Deprecations.warn(
        "#{input.inspect} is mutable."\
        " Be careful: types will return the same instance of the fallback"\
        " value every time. Call `.freeze` when setting the fallback."\
        "\n#{where}",
        tag: :"dry-types"
      )
    end

    if valid?(input)
      optional.append { |v| v.nil? ? input : v }.default(input)
    else
      raise Dry::Types::ConstraintError.new("fallback value #{input.inspect} violates constraints", input)
    end
  end
end
```

As you see it just appends a new constructor. But also makes the type `default` to be properly handled by dry-struct.
Also, I need to call `optional` first because the constructor gets simply replaced on Constructor types

## Examples

`Types::Params::Integer.fallback(0)` gives us the same effect as was achieved with `Types::Params::Integer.default(0)` before.

We can think about the details of implementation but I believe this feature can be considered for the next version as it is useful.

solnic:
@pyromaniac could you explain how the behavior of `default` actually changed so that it breaks for you now?

pyromaniac:
It didn't fallback to the default when nil was passed explicitly. In general, we lack value normalization in dry-types I believe. Something that adjust values after coercion. Normalization, for example, can be used to turn empty string to nil as well. This fallback proposal is simply a nerfed normalization.

flash-gordon:
I'm thinking about having `Wrap` types or something like that, it would be more general than constructors:
```ruby
Types::Integer.wrap do |type, input, &fallback|
  type.(input, &fallback) + 1
end
```
Similar to how rack middleware works. I'm not quite sure it solves "all the problems" but it does add some flexibility. WDYT?

pyromaniac:
This would be actually perfect. I'll be able to build my own fallback type basing on this. Not sure if I got the interface right though. Why do we need those arguments? Wouldn't simple 

```ruby
[...]

PR #410, opened 2020-12-15
Wrapping constructor types

flash-gordon:
This adds support for constructor types that can hook into "type application" by taking control when the block has two parameters:
```ruby
# This type parses input integer value and then adds 1 to the result
Month = Dry::Types["coercible.integer"].constructor do |input, type|
  type.(input) + 1
end
```
Earlier, this could be done with just `.constructor { |input| + 1 }`, however, having control over `type` allows for implementing various fallback strategies:
```ruby
Age = Dry::Types["coercible.integer"].constructor do |input, type|
  type.(input) { 18 }
end
Age.('17') # => 17
Age.('foo') # => 18
```
These kinds of constructor blocks also support `prepending` and `appending`. Generally, this interface is akin to the rack's middleware.

A word of warning: constructor types can get tricky, this is especially true for wrapping constructors. For instance, calling `.lax` on such a type would be a no-op since it's assumed you handle failing cases manually (it's actually required for the use case in dry-schema).

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -127,7 +127,7 @@ module Dry
       #
       # @api public
       def constructor(constructor = nil, **options, &block)
-        constructor_type.new(with(**options), fn: constructor || block)
+        constructor_type[with(**options), fn: constructor || block]
       end
       alias_method :append, :constructor
       alias_method :prepend, :constructor
--- a/lib/dry/types/constrained.rb
+++ b/lib/dry/types/constrained.rb
@@ -29,11 +29,9 @@ module Dry
         @rule = options.fetch(:rule)
       end
 
-      # @api private
-      #
       # @return [Object]
       #
-      # @api public
+      # @api private
       def call_unsafe(input)
         result = rule.(input)
 
@@ -44,11 +42,9 @@ module Dry
         end
       end
 
-      # @api private
-      #
       # @return [Object]
       #
-      # @api public
+      # @api private
       def call_safe(input, &block)
         if rule[input]
           type.call_safe(input, &block)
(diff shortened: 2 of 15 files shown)


solnic:
@flash-gordon what problems will this solve?

flash-gordon:
@solnic [this](https://github.com/dry-rb/dry-types/issues/407) and [that](https://discourse.dry-rb.org/t/something-like-lax-but-return-nil-instead-of-input/1168/6) and possibly a bunch of others not yet reported :)

solnic:
@flash-gordon ah OK, I had a hunch it was about this but wanted to make sure I understand it correctly. It's great you're working on it!

pyromaniac (commented):
I've got a little concern regarding the implementation. Should we really add additional responsibilities/features to the `constructor` method? I believe, it might affect the API evolution in the future. Why don't be explicit and add one more method to the API? Like `wrap`.

https://github.com/dry-rb/dry-types/pull/410/files#diff-30c2a0c967a56a3d5f181b895d0b488c78bb1c61b1181d0f823b2ade756efcd4R146 - this particular line is especially worrisome to me, we are basically starting to look for hacks in order to make it work. And despite it might be a source of quirks and unintended behavior, this also might not be good in a long run as it will be more and more difficult to change this logic in the future.

flash-gordon:
@pyromaniac adding a new type won't bring anything to the table. Adding new types is hard because it touches a lot of places, even with this one I spent > 10 hrs aligning all the things. Making a new type would be even more work. `Function` is a private API, it can be re-done as needed and it's a really minor concern.

pyromaniac:
I'm pretty sure you know what you are doing. Just expressing my general concerns basing on my experience. Ultimately, a new type going to be very similar to the `Constructor` type so I believe you made it right. It just looks a bit hacky you know :) Does not fit the existing concept naturally. But yeah, it might be addresses during the further refactorings.

solnic (approved):
👏🏻 👏🏻 👏🏻 🙇🏻

PR #411, opened 2020-12-26
Types with fallback values

flash-gordon:
This adds the `.fallback` method that wraps a type with a fallback value or a block. It's different from `.default` since default types make sense only for missing/absent values (e.g. missing keys in schema input). `.fallback` is more general:

```ruby
age = Types::Integer.fallback(18)
age.(20) # => 20
age.("invalid") # => 18
```

In some cases, `.default` and `.fallback` can be used interchangeably but I advise against this since it doesn't always use as you would expect:
```ruby
age.() # => 18
Types::Hash.schema(age: age).({}) # => :age is missing in Hash input
```

Fallbacks are tested to be compatible with dry-schema, see https://github.com/dry-rb/dry-schema/pull/337/files#diff-753532781c2792be1bada750fdaca7f5f8bca2ac6021da8afea8916ae3a02f86

Under the hood, it uses wrapping constructors [added earlier](https://github.com/dry-rb/dry-types/pull/410) so it just utilizes an existing mechanism.

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -59,7 +59,7 @@ module Dry
       # Turn a type into a type with a default value
       #
       # @param [Object] input
-      # @param [Hash] options
+      # @option [Boolean] shared Whether it's safe to share the value across type applications
       # @param [#call,nil] block
       #
       # @raise [ConstraintError]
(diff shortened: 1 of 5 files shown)


lib/dry/types/builder.rb
```diff
@@ -133,6 +133,43 @@ def constructor(constructor = nil, **options, &block)
...
+            ' Be careful: types will return the same instance of the fallback'\
+            ' value every time. Call `.freeze` when setting the fallback'\
+            ' or pass `shared: true` to discard this warning.'\
+            "\n#{where}",
+            tag: :'dry-types'
+          )
+        end
+
+        constructor do |input, type, &block|
+          type.(input) do |output = input|
```
pyromaniac, line 164: `output = input` this one looks so cryptic. Took some time for me to grasp it.

PR #412, opened 2021-01-16
Add params.string

### LINKED ISSUE #403: Include `Types::Params::String`
When including types using `params` as the default

```ruby
module Types
  include Dry.Types(default: :params)
end
```

for use in form objects for example, it would be nice to be able to define an attribute in a struct as being a regular string

```ruby
class MyStruct < Dry::Struct
  attribute :some_integer_param, Types::Integer
  attribute :name, Types::String
end
```

Currently this doesn't work and you instead need to define the `name` attribute with `Types::Nominal::String` which just feels a bit off.

I currently work around this by just defining my own `String` inside the `Types` module, but I was just wondering whether it wouldn't make sense inside the `Params` namespace.

flash-gordon:
Logically, it's no different from strict.string. It seems like dry-schema is fine with the change, I'll do a few more tests, though.

### DIFF
--- a/lib/dry/types/params.rb
+++ b/lib/dry/types/params.rb
@@ -56,6 +56,8 @@ module Dry
       self["nominal.symbol"].constructor(Coercions::Params.method(:to_symbol))
     end
 
+    register("params.string", self["string"])
+
     COERCIBLE.each_key do |name|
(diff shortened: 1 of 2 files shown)


flash-gordon:
CI failures are unrelated

PR #413, opened 2021-01-16
Add :builders extension with custom type constructors

flash-gordon:
This kinda extends our API and we could use, say, refinements instead. But as far as I can see it should be safe. @solnic @timriley @ianks thoughts?

Alternatively, we could add an API for registering custom constructors:
```ruby
Dry::Types.load_extensions(:builders)
Dry::Types.define_builder(:or_nil) { |type| type.optional.fallback(nil) } 
```

ref #409

### DIFF
--- a/lib/dry/types/extensions.rb
+++ b/lib/dry/types/extensions.rb
@@ -7,3 +7,7 @@ end
 Dry::Types.register_extension(:monads) do
   require "dry/types/extensions/monads"
 end
+
+Dry::Types.register_extension(:builders) do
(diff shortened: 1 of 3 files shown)


solnic:
@flash-gordon I like your idea with `define_builder` better; this would cover a very common scenario and give people a 1st-class feature versus patching internals of the library. It would also give *us* more flexibility as the feature could evolve and we wouldn't have to be tied to the `Builder` module.

flash-gordon:
Closed in favor of #415

PR #415, opened 2021-01-17
Support for custom builder methods

flash-gordon:
Given we have a stable and flexible API for building types I think it's relatively safe to give users a way to define custom type constructors. An even "safer" way would be using refinements but it doesn't look it's worth it.

ref #413, #409


solnic (approved):
I love this! Amazing job @flash-gordon (here and other related PRs) 👏🏻 🚀 🎉

PR #418, opened 2021-02-15
Add missing dry/core/equalizer requires

timriley:
Putting these in any file where we refer to Dry::Equalizer should prevent `undefined method 'Equalizer' for Dry:Module` occurring, regardless of the order of requires.

@solnic @flash-gordon Is there any reason these weren't here before? We get the error above today in CI when upgrading a work project to 1.5.0.

### DIFF
--- a/lib/dry/types/constrained.rb
+++ b/lib/dry/types/constrained.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "dry/core/equalizer"
 require "dry/types/decorator"
 require "dry/types/constraints"
 require "dry/types/constrained/coercible"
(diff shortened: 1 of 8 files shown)


solnic:
> @solnic @flash-gordon Is there any reason these weren't here before?

no, this was missed by an accident and now it's been revealed due to moving to dry-core

solnic (approved):

flash-gordon (approved):
I am wondering why it worked before. @timriley how do you `require` dry-types?

timriley:
Thanks folks :) Am releasing this as 1.5.1 now. I confirmed it fixed our build issue earlier today. I'll check back to work out how we're loading dry-types tomorrow. It's within a Rails app.

solnic:
@timriley FWIW - I don't think this was anything unexpected, it just happened that in your case files were loaded previously in such an order that the equalizer happened to be already loaded before loading dry-types.

PR #420, opened 2021-03-26
Optimize PredicateRegistry for Ruby 2.7+

casperisfine:
2.7 introduced `UnboundMethod#bind_call` which is much faster than
doing `Method.bind(receiver).call`.

And even on older rubies caching `Kernel.instance_method(:respond_to?)`
in a constant will save quite a lot of extra costly allocations.

### DIFF
--- a/lib/dry/types/predicate_registry.rb
+++ b/lib/dry/types/predicate_registry.rb
@@ -14,20 +14,35 @@ module Dry
       # @api private
       attr_reader :has_predicate
 
-      # @api private
-      def initialize(predicates = Logic::Predicates)
-        @predicates = predicates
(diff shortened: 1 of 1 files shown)


flash-gordon:
@casperisfine could you show that this optimization makes a difference for a typical dry-types usage? I don't think the registry is queried in "runtime" (when types are already defined) but I might be wrong.

casperisfine:
> I don't think the registry is queried in "runtime"

I should have specified that I found this in our boot profiling reports. Boot time is important to us so we have daily profiling of it.

flash-gordon:
Ok, fair enough. I think we could a cache or something, this could be even faster but it'll require benchmarking. I'll merge it as it is.

PR #424, opened 2021-07-30
Add missing test for a default with Undefined

waiting-for-dev:
When Dry::Core::Constants::Undefined is given to a type with default,
the default value is returned. The test for this behavior was missing.

### DIFF
--- a/spec/dry/types/default_spec.rb
+++ b/spec/dry/types/default_spec.rb
@@ -10,6 +10,10 @@ RSpec.describe Dry::Types::Builder, "#default" do
       expect(type[]).to eql("foo")
     end
 
+    it "returns default value when Undefined is passed" do
+      expect(type[Undefined]).to eql("foo")
+    end
+
     it "aliases #[] as #call" do
       expect(type.call).to eql("foo")
     end


flash-gordon:
It was kinda intentional, why would you provide `Undefined` explicitly? I don't feel like acknowledging "this is a public interface".

waiting-for-dev:
> It was kinda intentional, why would you provide `Undefined` explicitly? I don't feel like acknowledging "this is a public interface".

Wouldn't we consider it a regression if it returned `Undefined` instead? Probably an end-user won't provide `Undefined` directly, but if you use `dry-types` as a dependency while you're developing another library, you could rely on that behavior. So I think we should consider it as a public interface, too.

FWIW, I wanted to check that while debugging hanami/hanami-2-application-template#43

flash-gordon:
FWIW current interface won't change but I'd say the public interface is `type[*(condition(value) ? [] : [value])]` rather than passing `Undefined` explicitly. Providing `Undefined` as an input is a trick that shouldn't exist because it "exploits" implementation details. I'll merge this spec and add a comment just to make it clear in the code. But maybe I'll reconsider this in future, it's entirely possible :)

waiting-for-dev:
Thanks, @flash-gordon. I get your point. Yeah, we can change it in the future :slightly_smiling_face:

flash-gordon:
you know what? We already have a spec for `.valid?` https://github.com/dry-rb/dry-types/blob/148fe7f5a07be7064f7d58c33b15298b39c27a0b/spec/dry/types/default_spec.rb#L172
So adding more hassle around it is senseless, I'll just leave it without comments. My only point is `Undefined` has a special treatment so people shouldn't "hack" types with it. At the same time, I've no idea what to hack here, just don't make silly things :)

PR #427, opened 2021-09-27
Implement SchemaError#value and SchemaError#key

oleander:
This PR implements `SchemaError#key` & `SchemaError#value`. It makes `SchemaError`s interface more in line with `MissingKeyError`, `UnknownKeysError` & `ConstraintError` which have (most) of their instance variables accessible to the public.


flash-gordon:
thanks!

PR #430, opened 2021-12-22
Update example to use new Dry.Types() syntax

VikiAnn:
Tiny docs change, just ran into a deprecation error when attempting to use the example code.

### DIFF
--- a/docsite/source/index.html.md
+++ b/docsite/source/index.html.md
@@ -28,7 +28,7 @@ require 'dry-types'
 require 'dry-struct'
 
 module Types
-  include Dry::Types.module
+  include Dry.Types()
 end
 
 class User < Dry::Struct


VikiAnn:
Hmm, so it occurred to me to run a search for any other instances of the old syntax, and I was surprised to find no examples at all - including the one in this PR. Looks like the docs published to the website are from a very old version of the gem. That's probably the actual fix needed, not this PR.

solnic:
Hey, thanks for the PR - docs are versioned, so this: https://dry-rb.org/gems/dry-types/0.15/ points to dry-types 0.15.x docs and back then we had this syntax, so docs are correct.

As you didn't notice this, I suspect we could do a better job at displaying some info about the current version of docs you're reading. Maybe even add some extra notice if the docs you're reading are **old** 🙂 WDYT?

VikiAnn:
> As you didn't notice this, I suspect we could do a better job at displaying some info about the current version of docs you're reading. Maybe even add some extra notice if the docs you're reading are **old** 🙂 WDYT?

That sounds like it would be helpful, yes :D Thank you!

solnic:
I reported an issue about it so that we won't forget https://github.com/dry-rb/dry-rb.org/issues/388

PR #432, opened 2022-01-04
Update docs to recommend upfront module definition (i.e. `Types = Dry.Types()`)

### LINKED ISSUE #422: Params namespace disapearing
## Describe the bug

It is not possible to add new type definition in the Params namespace of included Types() using the following construction,
(it is certainly the same for Coercible, ...)

## To Reproduce

~~~ruby
module Types
  include Dry.Types()
  
  module Params
    PageSize = Types::Params::Integer
  end
end
~~~

Getting:
~~~
(irb):39:in `module:Params’: uninitialized constant Types::Params::Integer (NameError)
~~~

timriley:
As agreed, this is our recommendation to resolve the issue raised in #422.

Most of our code examples depend on a `::Types` module having been already created, so to make sure those examples are as clear as possible, I've also added `Types = Dry.Types()` in front of any grouping of code examples.

Made a few other very minor adjustments along the way, too.

Resolves #422


timriley:
@solnic @flash-gordon — are you happy with this? If so, I'll merge this and actually make sure we have a 1.5 version of the docs published at dry-rb.org.

docsite/source/getting-started.html.md
```diff
@@ -4,24 +4,22 @@ layout: gem-single
...
+    Types = Dry.Types()
     ```
-   
-2. Reload the environment, & type `Types::Coercible::String` in the ruby console to confirm it worked:
+
+2. Reload the environment, & enter `Types::Coercible::String` in your ruby console to confirm it worked:
 
     ``` ruby
     Types::Coercible::String
     # => #<Dry::Types::Constructor type=#<Dry::Types::Definition primitive=String options={}>>
```
radar, line 19: This now appears as:

```
=> #<Dry::Types[Constructor<Nominal<String> fn=Kernel.String>]>
```

timriley: Oh thanks, let me update that now.

docsite/source/built-in-types.html.md
```diff
@@ -15,7 +15,7 @@ Built-in types are grouped under 6 categories:
 
 ### Categories
 
-Assuming you included `Dry::Types` ([see instructions](docs::getting-started)) in a module called `Types`:
+Assuming you've defined your own `Types` module ([see instructions](docs::getting-started)):
```
adam12, line 18: How do you feel about changing the assertion to one that would be more indicative that they _should_ do this, rather than conditional that they _might_ have.

```suggestion
After you've defined your own `Types` module ([see instructions](docs::getting-started)):
```

flash-gordon (approved):

solnic (approved):

PR #433, opened 2022-01-04
Update README.md badge [ci skip]

seuros:
The workflow name is case sensitive


solnic:
Thanks but I can't merge it in because we keep things in sync using a template repo here: https://github.com/dry-rb/template-gem/blob/main/README.md.erb#L11

Feel free to open a PR there and the fix will be propagated in all dry-rb repos.

seuros:
Interesting template usage.

Here is the PR https://github.com/dry-rb/template-gem/pull/1. 
Thanks

solnic:
> Interesting template usage.

yeah it's cool because look what just happened after I merged your PR in the template repo:

Screenshot 2022-01-05 at 11 08 56

PR #444, opened 2022-08-10
Zeitwerk fixes

robhanlon22:
When using `Dry::Types.loader.eager_load`, Zeitwerk raises more than one error:

```
Zeitwerk::NameError:
  expected file .../dry-types/lib/dry/types/compat.rb to define constant Dry::Types::Compat, but didn't

NameError: uninitialized constant Dry::Types::Extensions
  .../dry-core/lib/dry/core/deprecations.rb:227:in `block (2 levels) in deprecate_constant'

NameError: uninitialized constant RSpec
  from .../dry-types/lib/dry/types/spec/types.rb:3:in `<top (required)>'
```

Add some ignores to prevent this error; add a test that ensures eager loading works.

### DIFF
--- a/spec/dry/types_spec.rb
+++ b/spec/dry/types_spec.rb
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 RSpec.describe Dry::Types do
+  describe ".loader" do
+    it "can eagerly load this library" do
+      Dry::Types.loader.eager_load
+    ensure
+      Dry::Types.loader.unload
+    end
+  end
+
   describe ".register" do
     it "registers a new type constructor" do
       module Test
--- a/lib/dry/types.rb
+++ b/lib/dry/types.rb
@@ -36,6 +36,8 @@ module Dry
         loader.push_dir(root)
         loader.ignore(
           "#{root}/dry-types.rb",
+          "#{root}/dry/types/extensions",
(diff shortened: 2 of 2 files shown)


robhanlon22:
An alternative would be to remove `compat.rb`, since it's empty; but, with a name like `compat.rb` I'm sure there's a reason it exists :D

flash-gordon:
There is, it's probably kept to avoid a breaking change on a minor version update. It's a guess, though.

Thank you!

spec/dry/types_spec.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 RSpec.describe Dry::Types do
+  describe ".loader" do
+    it "can eagerly load this library" do
+      Dry::Types.loader.eager_load
```
robhanlon22, line 6: @flash-gordon this might be nice to add as a "standard" test to each Zeitwerkified library; it also might make sense to shell out to a subprocess that does not load RSpec, since `types.rb` _may_ have successfully loaded in a test context (although Zeitwerk may have complained about `types.rb` not defining a Types class).

solnic:
`compat.rb` was added prior 1.0 when we removed a bunch of build in types (form stuff). It should be OK to remove this file now.

PR #445, opened 2022-08-20
Intersection (&) support redux

robhanlon22:
dry-schema and dry-logic both support &. Adding support for & in
dry-types allows for a more expressive type system...

```
module Types
  Callable = Interface(:call)
  Procable = Interface(:to_proc)
  Function = Callable & Procable
end
```

...and greatly simplifies value coercion in dry-schema, where a custom
types merge attempts to simulate an intersection type.

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -5,6 +5,7 @@ module Dry
     # Common API for building types and composition
     #
     # @api public
+    # rubocop:disable Metrics/ModuleLength
     module Builder
       include Dry::Core::Constants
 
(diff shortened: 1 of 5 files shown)


robhanlon22:
@flash-gordon any thoughts on this? I reopened and finished this work because of the comments in the original intersection support PR (#442) and that it would make downstream `&` composition in `dry-schema` less hacky. There's also a matching implication support PR (#446) as well as a composition refactor PR (#447) as well that could make this support generally more holistic.

lib/dry/types/builder.rb
```diff
@@ -30,8 +31,18 @@ def constructor_type
...
+
+      # Compose two types into an Intersection type
+      #
+      # @param [Type] other
+      #
+      # @return [Intersection, Intersection::Constrained]
+      #
+      # @api private
+      def &(other)
+        compose(other, Intersection)
```
robhanlon22, line 45: This is actually a circular dependency! Any thoughts on how we could fix this issue?

robhanlon22:
Enjoy your trip @flash-gordon !

flash-gordon:
I'll be merging the PRs, I'll post-review them

flash-gordon:
They all look good at the first glance btw

PR #453, opened 2023-02-11
Implement Transition composition type

estum:
It is a composition type, where the left type is an intermediary one and
the right type is a resulting.

It differs from an `Implication` by an input for the right type: instead of bypassing the original input, it will pass a coerced result of the left type. The same effect is possible with a `Constructor` but `Transition` allows to keep a transitive object as type instead of function.

When the left type if failed it bypasses input value to the right type.

The meta is always stored and get from the right-hand type, because the left type is not the target.

Here is a real-world example, that I use, note a combination with implication:

```ruby
Callable = Types.Interface(:call)
Aref = Types.Interface(:[])
CallableAref = Callable <= (Aref > Types.Constructor(Proc, proc(&:[]).curry(2)))
```

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -56,6 +56,28 @@ module Dry
         compose(other, Implication)
       end
 
+      # Compose two types into an Transition type
+      #
+      # @param [Type] right resulting type
+      #
+      # @return [Transition, Transition::Constrained]
+      #
+      # @api public
+      def >=(right)
+        compose(right, Transition)
+      end
+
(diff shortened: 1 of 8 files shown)


flash-gordon:
I like the idea but haven't looked into the implementation yet. However, I think `<=` is redundant. Shouldn't `>=` suffice?

flash-gordon:
By the way, not to discourage you but pls submit a question about adding new features on the forum first. It would be a shame if you spend a lot of your time on something that is rejected (or requires a lot of modification) for whatever reason.

PR #454, opened 2023-02-15
Fixed duplicate key warning

klobuczek:
Without this change, our logs are getting filled with these ruby warnings:
```
warning: key :fn is duplicated and overwritten on line 27
```

### DIFF
--- a/lib/dry/types/constructor.rb
+++ b/lib/dry/types/constructor.rb
@@ -24,7 +24,7 @@ module Dry
       # @api public
       def self.new(input, **options, &block)
         type = input.is_a?(Builder) ? input : Nominal.new(input)
-        super(type, **options, fn: Function[options.fetch(:fn, block)])
+        super(type, **options.merge(fn: Function[options.fetch(:fn, block)]))
       end
 
       # @param [Builder, Object] input


flash-gordon:
hm, where does this warning happen? I don't see it locally, that's strange. What is your ruby version?

klobuczek:
It happens for us on jruby-9.4.1.0 only which is strange. But you can see the warning on any ruby:
```
3.2.0 :001 > def m(a:); end
 => :m
3.2.0 :002 > m(a: 1, **{a: 2})
(irb):2: warning: key :a is duplicated and overwritten on line 2
 => nil
```
The `new` function is designed for this warning to happen. If `options` has the `:fn` key then you will always get the warning. But if `options` should never have `:fn` then why even try to `fetch(:fn)`? In either case, this needs some correction. 
The question remains why potentially for the first time in jruby-9.4.1.0 the function is called with `:fn` key in `options`?

flash-gordon:
I think this is a bug in jruby. Cruby only issues warnings when literals are duplicated which makes perfect sense. When you use `**` on a hash it doesn't make sense to filter out duplicates, consider something like this:
```ruby
meth(**kwargs1, **kwargs2, **kwargs3, foo: 123)
```
It would be *very* tedious to check all hashes for having the `:foo` key for no good reason. I'll fix this particular case but generally it should be reported and fixed in jruby.

Issue #28, opened 2016-02-07
File type


kirs:
It would be great to have the built-in `File` type in `dry-data` and `dry-validation`.
Since `dry-validation` is a solution for request parameters / form validation, `File` a kind of must have type for forms.

@solnic what's your opinion?

solnic:
@kirs yes, it'm open to add more types like `File`. I'm gonna hold off a little bit though, as I'm planning to refine behavior of `Type` objects for the next version. It should not take too long, shooting for early March, maybe earlier.

kirs:
Does it mean that the public API will change?

solnic:
@kirs no, it's gonna be an internal clean up with refined internals and an improved API for defining custom types. I need to make sure that more complex types like sums, optionals, default-values can work correctly without too much internal complexity. Although it's hard to say which API is public, the lib is not stable in general (< 1.0). Can you tell which APIs you depend on already?

kirs:
I don't depend on any particular API, just wondering about your plans about it.

jeremyf:
:+1: for this feature; I'd be up for working on the `File` type.

solnic:
@jeremyf still up for the task? :)

PR #455, opened 2023-04-17
Add File built-in type

SalvatoreT:
I had the same thought as the folks in this discussion, so I took a quick swing at the task.

https://github.com/dry-rb/dry-types/issues/28

### DIFF
--- a/lib/dry/types/coercions.rb
+++ b/lib/dry/types/coercions.rb
@@ -77,6 +77,29 @@ module Dry
         end
       end
 
+      # @param [#to_str, Object] input
+      #
+      # @return [File, Object]
+      #
+      # @see File.open
+      #
+      # @api public
+      def to_file(input, &block)
+        if input.is_a?(::File)
+          input
+        elsif input.respond_to?(:to_str)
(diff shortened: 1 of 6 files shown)


SalvatoreT:
Hey @solnic, is similar to what you had imagined for a built-in File type?

SalvatoreT:
Can I please get 👀 on this @timriley @flash-gordon?

lib/dry/types/coercions.rb
```diff
@@ -77,6 +77,29 @@ def to_time(input, &block)
...
+            ::File.open(input)
+          rescue ArgumentError, Errno::ENOENT => e
+            CoercionError.handle(e, &block)
+          end
+        elsif block_given?
+          yield
+        else
+          raise CoercionError, "#{input.inspect} is not a string"
+        end
+      end
```
flash-gordon, line 101: Why would you need a coercion to a file? I don't think it's acceptable since it's a security concern. One thing is checking if a variable has a certain type, but opening files seems far beyond the responsibility of dry-types

PR #457, opened 2023-06-27
Register a boolean type as an alias of bool

radar:


flash-gordon:
On the one hand, it's convenient. On the other,  your code may now use two different names for the same thing and IMO it doesn't help. I prefer raising a friendly error when someone reaches for `boolean` instead of `bool`. WDYT?

PR #460, opened 2023-08-31
Give Enum types a name with meaningful information

krainboltgreene:
The name of enum types don't display their state information.

Given: `Strict::String.enum(['pending', 'active', 'frozen', 'in_review', 'banned', 'incomplete']).optional.name`
Before: `String | NilClass`
After: `String[pending|active|frozen|in_review|banned|incomplete] | NilClass`

### DIFF
--- a/lib/dry/types/enum.rb
+++ b/lib/dry/types/enum.rb
@@ -33,6 +33,10 @@ module Dry
         freeze
       end
 
+      def name
+        "#{super}[#{values.join('|')}]"
+      end
+
       # @return [Object]
       #
       # @api private


krainboltgreene:
I can add tests/docs after I get some feedback (or none if this isn't desired).

flash-gordon:
This makes perfect sense to me.

krainboltgreene:
So I just used this as a patch in my own project and I'm noticing that `Strict::String.enum([:a, :b, :c])['a']` raises a constraint error. First off I'm not sure why that type annotation is allowed if they're not the same? Maybe I should write a patch for that as well? And second this seems to imply we should call `inspect()` on the values.

flash-gordon:
> So I just used this as a patch in my own project and I'm noticing that Strict::String.enum([:a, :b, :c])['a'] raises a constraint error. First off I'm not sure why that type annotation is allowed if they're not the same?  Maybe I should write a patch for that as well? 

Can you file a separate issue on this? It's not hard to fix, we just don't seem to check that passed values meet the constraints. We already fixed in for default types so that `Types::String.default(:a)` throws an error

> And second this seems to imply we should call inspect() on the values.
I agree, but I suggest skipping `.inspect` on strings. The output would be nicer.

PR #467, opened 2024-01-25
Added documentation for intersection types

emptyflask:
I was working on a project and wanted this functionality, happened to just try using `&` and it actually worked -- but discovered that this was still an undocumented feature.

So, I wrote up a short page about it, using an example from my real world project.

I also arranged sum & intersection pages under a "Combining Types" section, and placed it after constraints and hash schemas, since this sort of builds on top of those ideas. I didn't see a way to actually test the docsite generation, so hopefully it looks good.

### DIFF
--- a/docsite/source/combining-types.html.md
+++ b/docsite/source/combining-types.html.md
@@ -1,0 +1,12 @@
+---
+title: Combining Types
+layout: gem-single
+name: dry-types
+sections:
+  - intersection
+  - sum
+order: 11
+---
+
+Types can be combined to create new types, using [intersection types](docs::combining-types/intersection)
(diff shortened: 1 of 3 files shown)


emptyflask:
Tagging @robhanlon22 in case you wanted to add anything, since you wrote the code.

solnic:
Thank you :) The docsite should be built via our devtools toolchain and end up automatically in dry-rb.org/gems/dry-types under Main version (then it will be included under the next released version too).

solnic:
Oops, unfortunately it didn't build correctly:

```
       error  docs/gems/dry-types/main/combining-types/index.html
section sum not found
```

emptyflask:
:grimacing: looks like I messed up the paths. I'll take a look and see if I can get it to build.

PR #468, opened 2024-02-06
fix docsite structure

emptyflask:
Follow-up to #467. This seems to work well now.

Sorry it took some time to get the whole docsite compilation toolchain working. I wasn't able to run middleman on the dry-rb.org project with any currently installable version of ruby, so I ended up converting the github CI action in that repo to a dockerfile.

### DIFF
--- a/docsite/source/combining-types.html.md
+++ b/docsite/source/combining-types.html.md
@@ -5,7 +5,6 @@ name: dry-types
 sections:
   - intersection
   - sum
-order: 11
 ---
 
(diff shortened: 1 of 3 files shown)


solnic:
woah thank you so much, I'm sorry it turned out to be so time consuming. We've been looking into moving to a different doc stack for this exact reason, but so far nobody had the time to work on migrating all of the things and we haven't yet decided which doc tool we'd like to use.

solnic:
OK it's live here https://dry-rb.org/gems/dry-types/main/combining-types/intersection/ :)

PR #474, opened 2024-07-15
Provide a consistent interface for specifying nullary constraints

### LINKED ISSUE #473: No obvious way to use nullary predicates with constraints
## Describe the bug

The documentation states that "Under the hood `dry-types` uses [dry-logic](https://dry-rb.org/gems/dry-logic) and all of its predicates are supported." However, there is no obvious way to use nullary predicates (predicates that don't take an argument), like `odd?`.

I previously raised this issue with the [community](https://discourse.dry-rb.org/t/using-boolean-like-predicates-as-constraints-in-dry-types/1807).

## To Reproduce

```ruby
odd = Types::Integer.constrained(:odd)

NoMethodError: undefined method 'map' for :odd:Symbol
from /home/damian/.gem/ruby/3.2.2/gems/dry-types-1.7.2/lib/dry/types/contraints.rb:15:in 'Rule'
```

## Expected behavior

There should be a conventional way to pass nullary predicates.

## Workarounds

After reviewing the `Rule` class method in `/lib/dry/types/constraints.rb`, I see that it only expects to receive keyword arguments that it can treat as a Hash and `map` over. This implied a workaround for using nullary predicates by including a dummy value.

```ruby
odd = Types::Integer.constrained(odd: nil)
=> #<Dry::Types[Constrained<Nominal<Integer> rule=[type?(Integer) AND odd?]>]>
```
@flash-gordon pointed out that you can also achieve the same workaround by passing an array of one or more nullary predicates.

```ruby
odd = Types::Integer.constrained([:odd])
=> #<Dry::Types[Constrained<Nominal<Integer> rule=[type?(Integer) AND odd?]>]>
```
However, this workaround does not allow users to combine nullary predicates with unary predicates, as can be done by explicitly passing dummy values.

```ruby
positivelyodd = Types::Integer.constrained(odd: nil, gt: 0)
=> #<Dry::Types[Constrained<Nominal<Integer> rule=[type?(Integer) AND odd? AND gt?(0)]>]>
```

## Offer

I would be willing to submit a pull request to correct this issue, but I didn't want to do so without a conversation on implementation.

It seems to me that the only way to handle this is to require nullary predicates to be grouped first, before the unary predicates that are passed as keyword arguments. This would allow us to glom the nullary predicates into an array, map this to a Hash with dummy values, and merge with the keyword argument Hash before processing as currently implemented.

If that sounds good, I would be happy to code it up. In addition, I would be willing to update the docsite with more discussion and examples of using `dry-logic` predicates as constraints in `dry-types`. (Partially addressing #390)

## Thank You

Thank you!

dcr8898:
Closes #473

Implements the following api for specifying nullary constraints (no additional argument needed, e.g., `:odd`) and unary constraints (additional argument needed, e.g., `gt: 18`).

```ruby
type.constrained(:odd) # new variant
type.constrained(:odd, :even) # new variant
type.constrained(:odd, gt: 200, lt: 300) # new variant
```

The above is covered by new tests and updated yardocs. If this api is acceptable, I will update the docsite accordingly.

## Addenda

The api also accepts the following for backward compatibility. However, I suggest deprecating this behavior as non-intuitive and dependent on a quirk of the former implementation (`#map`). Deprecating will allow for the removal of an otherwise unnecessary application of `Array#flatten` in processing arguments.

```ruby
type.constrained([:odd, :even]) # worked before, continue to support
type.constrained([:odd, :even], gt: 200, lt: 300) # new variant, supported as a side-effect
```
The following also work, although I don't believe they are covered by tests. The last two are no-ops.

```ruby
type.constrained # error, as before
type.constrained([]) # worked before, continue to support
type.constrained({}) # worked before, continue to support
```

Finally, the following also work as undocumented behavior.

```ruby
type.constrained({gteq: 0, lt: 100}) # undocumented, not encouraged
type.constrained({gteq: 0}, lt: 100) # undocumented, not encouraged
```

## Potential Issues

- Dry::Types::Sum.constrained (on line 89) does not double-splat the options Hash used to call `Constrained#constrained`, which means that it relies on the "undocumented" behavior above. Not a big deal, but I would consider adding `**` for clarity or, even better, using argument forwarding (`...`).

- Dry::Types::Builder.constrained does not implement the above api. Although this does not break current tests, this should probably be modified (with tests) to be consistent. Maybe just use argument forwarding here as well?

### DIFF
--- a/lib/dry/types/builder.rb
+++ b/lib/dry/types/builder.rb
@@ -72,8 +72,8 @@ module Dry
       # @return [Constrained]
       #
       # @api public
-      def constrained(options)
-        constrained_type.new(self, rule: Types.Rule(options))
+      def constrained(...)
+        constrained_type.new(self, rule: Types.Rule(...))
       end
 
       # Turn a type into a type with a default value
--- a/lib/dry/types/constrained.rb
+++ b/lib/dry/types/constrained.rb
@@ -75,8 +75,9 @@ module Dry
         end
       end
 
-      # @param [Hash] options
-      #   The options hash provided to {Types.Rule} and combined
+      # @param *nullary_rules [Array<Symbol>] a list of rules that do not require an additional argument (e.g., :odd)
+      # @param **unary_rules [Hash] a list of rules that require an additional argument (e.g., gt: 0)
+      #   The parameters are merger to create a rules hash provided to {Types.Rule} and combined
       #   using {&} with previous {#rule}
       #
       # @return [Constrained]
--- a/lib/dry/types/constrained.rb
+++ b/lib/dry/types/constrained.rb
@@ -84,8 +84,12 @@ module Dry
       # @see Dry::Logic::Operators#and
       #
       # @api public
-      def constrained(options)
-        with(rule: rule & Types.Rule(options))
+      def constrained(*nullary_rules, **unary_rules)
+        nullary_rules_hash = parse_arguments(nullary_rules)
+
+        rules = nullary_rules_hash.merge(unary_rules)
+
+        with(rule: rule & Types.Rule(rules))
       end
 
       # @return [true]
@@ -132,6 +136,17 @@ module Dry
       # @api private
       def decorate?(response)
         super || response.is_a?(Constructor)
+      end
+
+      # @param [Array] positional_args
+      #
+      # @return [Hash]
+      #
+      # @api private
+      def parse_arguments(positional_arguments)
+        return positional_arguments.first if positional_arguments.first.is_a?(::Hash)
+
+        positional_arguments.flatten.zip([]).to_h
       end
     end
   end
--- a/lib/dry/types/constrained.rb
+++ b/lib/dry/types/constrained.rb
@@ -75,8 +75,10 @@ module Dry
         end
       end
 
-      # @param *nullary_rules [Array<Symbol>] a list of rules that do not require an additional argument (e.g., :odd)
-      # @param **unary_rules [Hash] a list of rules that require an additional argument (e.g., gt: 0)
+      # @param *nullary_rules [Array<Symbol>] a list of rules that do not require an additional
+      #   argument (e.g., :odd)
+      # @param **unary_rules [Hash] a list of rules that require an additional argument
+      #   (e.g., gt: 0)
       #   The parameters are merger to create a rules hash provided to {Types.Rule} and combined
       #   using {&} with previous {#rule}
       #
--- a/lib/dry/types/sum.rb
+++ b/lib/dry/types/sum.rb
@@ -84,9 +84,9 @@ module Dry
       # @see Builder#constrained
       #
       # @api public
-      def constrained(options)
+      def constrained(...)
         if optional?
-          right.constrained(options).optional
+          right.constrained(...).optional
         else
           super
         end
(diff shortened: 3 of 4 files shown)


solnic:
```ruby
type.constrained([]) # worked before, continue to support
type.constrained({}) # worked before, continue to support
```

I think we should tweak it to raise an exception in 2.0

> Dry::Types::Sum.constrained (on line 89) does not double-splat the options Hash used to call Constrained#constrained, which means that it relies on the "undocumented" behavior above. Not a big deal, but I would consider adding ** for clarity or, even better, using argument forwarding (...).
> Dry::Types::Builder.constrained does not implement the above api. Although this does not break current tests, this should probably be modified (with tests) to be consistent. Maybe just use argument forwarding here as well?

It should be safe to use `...` there and it would be good to do it already in this PR :)

flash-gordon:
> I think we should tweak it to raise an exception in 2.0

Such calls can occur when dry-types is wrapped inside another lib.

dcr8898:
```ruby
type.constrained([]) # worked before, continue to support
type.constrained({}) # worked before, continue to support
```

I don't have an opinion either way on keeping this, but if it's important it should be covered by tests. I don't think it is.

dcr8898:
I have changed Dry::Types::Builder#constrained and Dry::Types::Sum#constrained to use argument forwarding. All tests remain green.

## Issues

- I don't know how to represent argument forwarding in Yardoc, so I didn't touch those.
- Passing an empty Array or Hash to `#constrained` doesn't *really* work (before or after the proposed changes).

### Yardoc

I found this example of argument forwarding in the codebase (for another invocation of `#constrained` no less!), but I wasn't sure if this is the canonical way to represent argument forwarding in Yardoc (as an `[Array]`). It looks suspect to me. :monocle_face: 

```ruby
      # Build a constrained type
      #
      # @param [Array] args see {Dry::Types::Builder#constrained}
      #
      # @return [Default]
      #
      # @api public
      def constrained(...)
        type.constrained(...).default(value)
      end
```
### Empty Array/Hash argument

Passing an empty Array or Hash to `#constrained` does not cause an explicit error either before or after the proposed changes. However, they both actually contain a subtle bug.

Entering `Types::Integer` in IRB produces:

```ruby
=> #<Dry::Types[Constrained<Nominal<Integer> rule=[type?(Integer)]>]>
```

Entering `Types::Integer.constrained([])` or `Types::Integer.constrained({})` should logically produce the same result. However, they actually both produce the following:

```ruby
=> #<Dry::Types[Constrained<Nominal<Integer> rule=[type?(Integer) AND ]>]>
```

That dangling `AND` does not seem to effect the functionality of the Type object, but it *does* break the AST parser. That is, the following works:

```ruby
irb(main):005:0> Types::Integer.to_ast
=> [:constrained, [[:nominal, [Integer, {}]], [:predicate, [:type?, [[:type, Integer], [:input, Undefined]]]]]]
```

But the following causes an error:

```ruby
irb(main):005:0> Types::Integer.constrained([]).to_ast
```

This bug stopped me from adding tests for empty Arrays/Hashes, since the test suite invokes `#to_ast` and then errors out.

## Path forward

Please let me know how you would like for me to proceed. I would be happy to update the Yardocs for the two methods I changed in whatever way you recommend.

As for the bug, do you want to address that in this PR, or separately? It might be as easy as adding a guard clause that checks for `rules.empty?` before calling `#with`, but I don't know the proper value to return from the guard clause.

Guidance is appreciated if you would like for me to investigate further.

flash-gordon:
@dcr8898 this looks good overall, I'll now merge it as it is and run my apps against it. If everything's fine, I'll return to the YARD nuances. Thank you for working on this! ❤️

flash-gordon:
So far, it works well, I'll look into YARD issues soon

PR #492, opened 2025-12-17
Upgrade bigdecimal to v4

rus-max:


timriley:
Unless we need specific v4 features of bigdecimal, I wonder if I better version constraint here is `">= 3.0"`?

Looking at the [changelog](https://github.com/ruby/bigdecimal/blob/master/CHANGES.md), it seems like mostly some deprecations and fixes, but this does seem like a breaking change, and I'd rather we not force it on anyone:

> BigDecimal#divmod return value changed to [Integer, BigDecimal]

timriley:
Nicely spotted, thanks @rus-max. Fixed with https://github.com/dry-rb/dry-types/commit/c4dc2ec432af2f694d245007b2f403d5f4a8fed9.

PR #493, opened 2026-02-03
Workaround for JRuby handling block default arguments

katafrakt:
This addresses https://github.com/jruby/jruby/issues/9208 by replacing `yield` with `block.call`, where the bug does not exist. From what I checked the performance difference is within a margin of error on MRI, although `block.call` seems to be slower in JRuby, so maybe we will want to revert it in the future, if if gets fixed.

The test suite for Dry Types did not catch this bug, but it manifests downstream - in Dry Schema na Hanami Controller. So perhaps we could release a new patch version after this?


timriley (approved):
Looks good!

> so maybe we will want to revert it in the future, if if gets fixed.

I wonder what will remind us to do this? Could we add at least one comment above somewhere in one of the changed lines of code, pointing to this PR? That way will have a reference back to all the places we changed because of this bug.

PR #494, opened 2026-02-12
Revert JRuby workaround

katafrakt:
Reverts cdff1829871839f41b93d4eb7904cd6cda9a4a22

The workaround was needed due to a bug in JRuby. The bug has since then been fixed and should be released in 10.0.4.0. Once this happens, this PR could be merged. So far I verified manually that the tests pass with current JRuby head.

### DIFF
--- a/lib/dry/types/array/member.rb
+++ b/lib/dry/types/array/member.rb
@@ -42,7 +42,7 @@ module Dry
         # @return [Array]
         #
         # @api private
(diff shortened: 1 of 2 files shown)


timriley (approved):

katafrakt:
Yes, I didn't expect it too. But JRuby got some great development speed recently. Kudos to the maintainers.

